PR #1, opened 2021-11-25
Design doc: modular_installation

eguzki:
As we are looking at leveraging the authconfig and ratelimit CRs directly. We have a natural gate to leveraging different functionality offered by kuadrant. So for modular installation we will ultimately look to enable pieces of Kuadrant via a Kuadrant operator and its CR. When instructed the Kuadrant operator would install a specific piece (authrino for example) and that in turn would make the API available. Initially we may also support this via installation on the CLI.

### DIFF
--- a/docs/design/modular_installation.md
+++ b/docs/design/modular_installation.md
@@ -1,0 +1,63 @@
+# Kuadrant Proposal - Modular Installation
+
+Kuadrant is developing a set of loosely coupled functionalities built directly on top of Kubernetes.
+Kuadrant aims to allow customers to just install, use and understand those functionalities they need.
+Currently, the installation tool of kuadrant, the [kuadrantctl CLI](https://github.com/Kuadrant/kuadrantctl),
+installs all or nothing. For instance, a customer not wanting to rate limit the API,
+should not have to install the rate limit component ([Limitador](https://github.com/Kuadrant/limitador)).
+
+## High Level Goals
+
+* Install only required components. Operate only required components.
+* Expose only the kuadrant API that has available implementation by the with the deployed components
+
+## Installation profiles
+
+The kuadrant installation mechanism should offer modular installation to enable pieces of kuadrant.
+Modular installation options should be feature oriented rather than deployment component oriented.
+Then, it is up to the installation tool to decide what components need to be deployed and how to
+configure it.
+
+Kuadrant features:
+
+* Multiple APIs bundled in a product
+* Service Discovery
+* TLS
+* Authentication (mTLS, API key, OIDC)
+* Rate Limiting (pre-auth, post-auth)
+
+Each feature, or part of it, is eligible to be included or excluded when installing kuadrant.
+
+Some profiles can be defined to group set of commonly required features. Naming the profiles
+allows the customer to easily express wanted installation configuration. Furthermore, profiles
+not only can be used to group a set of features, profiles can be used to define deployment options.
+
+| Name | Description |
+| ---  | --- |
+| **Minimal** | Minimal installation required to run an API without any protection, analytics or API management |
+| **Authenticated** | MTLS, API key and OIDC authentication features |
+| **RateLimit** | Basic rate limit (only pre-auth rate limit) features |
+| **Full** | Full featured kuadrant installation |
+| **Development** | Full featured kuadrant for development purposes. Development deployment mode: debug log level, no cache, tracing |
+
+A kuadrant operator, together with a design of a kuadrant CRD is desired.
+Not only for kuadrant installation, but also for lifecycle management.
+Additionally, the [kuadrantctl](https://github.com/Kuadrant/kuadrantctl) CLI tool can also
+be used to install kuadrant in a profile basis. The CLI tool may also accept, optionally, the kuadrant
+custom resource as input.
+
+## Available API
+
+The API exposed to the customer is defined in terms of k8s Custom Resource Definitions.
+Kuadrant provides multiple CRDs to configure kuadrant API management.
+Often, a kuadrant feature's fully definition spans multiple CRDs, so the available API cannot be
+accurately defined with a set of CRD. In other words, there is not such a mapping function that maps
+each feature with a CRD.
+
+The [kuadrant-controller](https://github.com/Kuadrant/kuadrant-controller) will be aware of the
+installation profile via env vars (or command line params). With that information,
+the kuadrant controller will reject any configuration object (all the object at once as partial rejection is not supported)
+when that configuration requires something (could be a component or deployment option) from kuadrant
+that is not enabled as part of the installation process.
+
+The customer will be only able to configure available features, nothing more.
--- a/docs/design/modular_installation.md
+++ b/docs/design/modular_installation.md
@@ -2,29 +2,33 @@
 
 Kuadrant is developing a set of loosely coupled functionalities built directly on top of Kubernetes.
 Kuadrant aims to allow customers to just install, use and understand those functionalities they need.
+
+## Problem Statement
+
 Currently, the installation tool of kuadrant, the [kuadrantctl CLI](https://github.com/Kuadrant/kuadrantctl),
-installs all or nothing. For instance, a customer not wanting to rate limit the API,
-should not have to install the rate limit component ([Limitador](https://github.com/Kuadrant/limitador)).
+installs all or nothing. Installing more than the customer needs adds unneeded complexity and operational effort.
+For example, if a customer is looking for rate limiting and not interested in authentication functionality,
+then the customer should be able to just install and run that part of Kuadrant.
 
 ## High Level Goals
 
 * Install only required components. Operate only required components.
-* Expose only the kuadrant API that has available implementation by the with the deployed components
 
-## Installation profiles
+Reduce system complexity and operational effort to the minimum required.
+Components in this context make reference to deployments and running instances.
 
-The kuadrant installation mechanism should offer modular installation to enable pieces of kuadrant.
+* Expose only the activated functionalities
+
+A user of a partial Kuadrant install should not be confronted with data in custom resources that
+has no meaning or is not accessible in their partial Kuadrant install. The design of the kuadrant
+API should have this goal into account.
+
+## Proposed Solution
+
+The kuadrant installation mechanism should offer modular installation to enable/disable loosely coupled pieces of kuadrant.
 Modular installation options should be feature oriented rather than deployment component oriented.
 Then, it is up to the installation tool to decide what components need to be deployed and how to
 configure it.
-
-Kuadrant features:
-
-* Multiple APIs bundled in a product
-* Service Discovery
-* TLS
-* Authentication (mTLS, API key, OIDC)
-* Rate Limiting (pre-auth, post-auth)
 
 Each feature, or part of it, is eligible to be included or excluded when installing kuadrant.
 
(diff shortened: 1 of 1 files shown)


docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+
+| Name | Description |
+| ---  | --- |
+| **Minimal** | Minimal installation required to run an API without any protection, analytics or API management |
+| **Authenticated** | MTLS, API key and OIDC authentication features |
+| **RateLimit** | Basic rate limit (only pre-auth rate limit) features |
+| **Full** | Full featured kuadrant installation |
+| **Development** | Full featured kuadrant for development purposes. Development deployment mode: debug log level, no cache, tracing |
+
+A kuadrant operator, together with a design of a kuadrant CRD is desired.
```
maleck13, line 43: Do we want to propose a CRD design here or elsewhere?

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+Not only for kuadrant installation, but also for lifecycle management.
+Additionally, the [kuadrantctl](https://github.com/Kuadrant/kuadrantctl) CLI tool can also
+be used to install kuadrant in a profile basis. The CLI tool may also accept, optionally, the kuadrant
+custom resource as input.
+
+## Available API
+
+The API exposed to the customer is defined in terms of k8s Custom Resource Definitions.
+Kuadrant provides multiple CRDs to configure kuadrant API management.
+Often, a kuadrant feature's fully definition spans multiple CRDs, so the available API cannot be
```
maleck13, line 53: I am not sure I follow this part. Do you mean authenticated ratelimit for example?

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+accurately defined with a set of CRD. In other words, there is not such a mapping function that maps
+each feature with a CRD.
+
+The [kuadrant-controller](https://github.com/Kuadrant/kuadrant-controller) will be aware of the
+installation profile via env vars (or command line params). With that information,
+the kuadrant controller will reject any configuration object (all the object at once as partial rejection is not supported)
+when that configuration requires something (could be a component or deployment option) from kuadrant
+that is not enabled as part of the installation process.
+
+The customer will be only able to configure available features, nothing more.
```
maleck13, line 63: I think we can leave this out. I can imagine over time more things will be configurable via the operator CR (example a HA multi AZ configuration)

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+Kuadrant is developing a set of loosely coupled functionalities built directly on top of Kubernetes.
+Kuadrant aims to allow customers to just install, use and understand those functionalities they need.
+Currently, the installation tool of kuadrant, the [kuadrantctl CLI](https://github.com/Kuadrant/kuadrantctl),
+installs all or nothing. For instance, a customer not wanting to rate limit the API,
+should not have to install the rate limit component ([Limitador](https://github.com/Kuadrant/limitador)).
+
+## High Level Goals
+
+* Install only required components. Operate only required components.
+* Expose only the kuadrant API that has available implementation by the with the deployed components
```
mikenairn, line 12: Are you talking about the kubernetes API here? i.e. Don't create the CRDs for components of kuadrant that aren't installed?

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+
+Each feature, or part of it, is eligible to be included or excluded when installing kuadrant.
+
+Some profiles can be defined to group set of commonly required features. Naming the profiles
+allows the customer to easily express wanted installation configuration. Furthermore, profiles
+not only can be used to group a set of features, profiles can be used to define deployment options.
+
+| Name | Description |
+| ---  | --- |
+| **Minimal** | Minimal installation required to run an API without any protection, analytics or API management |
```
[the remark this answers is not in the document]
mikenairn: Seems reasonable, i assume there would be some minimal functionality that kuadrant provides and you add stuff on top?

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+| **Authenticated** | MTLS, API key and OIDC authentication features |
+| **RateLimit** | Basic rate limit (only pre-auth rate limit) features |
+| **Full** | Full featured kuadrant installation |
+| **Development** | Full featured kuadrant for development purposes. Development deployment mode: debug log level, no cache, tracing |
```
mikenairn, line 41: Does a development profile make sense? Seems different from the intention of the other profiles, and likely something you'd want to enabled on any configuration Minimum, Full etc..., or, is the intention that you can stack these profiles, have more than one enabled? In which case is this just a list of the features you can enable?

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+
+A kuadrant operator, together with a design of a kuadrant CRD is desired.
+Not only for kuadrant installation, but also for lifecycle management.
+Additionally, the [kuadrantctl](https://github.com/Kuadrant/kuadrantctl) CLI tool can also
+be used to install kuadrant in a profile basis. The CLI tool may also accept, optionally, the kuadrant
+custom resource as input.
```
mikenairn, line 47: kuadrantctl seems to currently hold the manifest for each kuadrant component (istio, limitador, authorino etc..), but the operator will also need those things. Is the intention here to have kuadrantctl install the operator and configure via the operators CR?

docs/design/modular_installation.md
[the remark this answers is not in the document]
maleck13: Yes I think by default the ingress would work but not rate limiting and auth

docs/design/modular_installation.md
maleck13: I think if we could integrate the ctl and opertor into a single project it could provide some advantages. One thing we would like to do is convert a Open API Spec into k8s kuadrant resources. Being able to do this as part of a CLI or as part of the operator would be valuable for giving end user the option for either

docs/design/modular_installation.md
guicassolato: Agreed. The flow for developers might be different, but it's a good thing that they interact with the system under the same deployment options and capabilities (same profile?) we offer to users.

As an example, one who's working on a feature to which the _RateLimit_ profile is supposed to suffice should not inadvertently count on features available on the _Auth_ profile, or else an automatic and perhaps unexpected requirement for using the _Full_ profile could emerge. At the very least, the boundaries between one profile and the other should be clear not only to users, but to developers as well and possibly most of all, IMHO.

That said, I'm guessing the intent for a _Development_ profile was to provide a "shortcut" to activate different log mode, disable caching, etc. Maybe the CLI can still offer that as a `--development` flag that can be used in combination with all the other profiles instead of alternatively to those?

docs/design/modular_installation.md
guicassolato: As a user myself, my own preference for the ctl would be for it to be as minimal as possible. Converting OAS into K8s resources is a good example of added value. Other than that, I personally would default to maintaining my sets of declarative YAMLs (whose structure I have to learn regardless), and using the k8s CLI, k8s API, rather than learning yet another layer of imperative commands.

Maybe it's only me.

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Available API
+
+The API exposed to the customer is defined in terms of k8s Custom Resource Definitions.
```
guicassolato, line 51: I see no mention to the other components (Authorino, Limitador) or to the fact that Kuadrant might after all depend on CRDs defined by those components and/or their operators. Would this be considered an implementation detail and not for the current level of this proposal yet or should we just go and state right away "Kuadrant operator will deploy operators X and Y, which will define other feature-specific CRDs..."?

docs/design/modular_installation.md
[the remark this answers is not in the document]
eguzki: Makes sense. I will add a note about being the default one.

docs/design/modular_installation.md
```diff
@@ -0,0 +1,63 @@
...
+## Installation profiles
+
+The kuadrant installation mechanism should offer modular installation to enable pieces of kuadrant.
+Modular installation options should be feature oriented rather than deployment component oriented.
+Then, it is up to the installation tool to decide what components need to be deployed and how to
+configure it.
+
+Kuadrant features:
+
+* Multiple APIs bundled in a product
```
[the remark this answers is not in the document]
eguzki: ok, let's remove them

docs/design/modular_installation.md
eguzki: I was inspired by the [istio installation profiles](https://istio.io/latest/docs/setup/additional-setup/config-profiles/). There is this `demo` profile that is what I meant by `Development`. Istio profiles are more complex and span a wider set of uses cases. For instance, you can have different profiles for the control plane and the data plane. Which, of course, can make sense here too.

I will re-think this again

docs/design/modular_installation.md
eguzki: Not here :)

docs/design/modular_installation.md
eguzki: For installation purposes, my proposal is that an operator is the first candidate and a CLI is optional. Mainly for covering lifecycle (upgrades) management needs. 

That said, the CLI may add value beyond the installation (and upgrades) features. For instance, as you say, creating k8s objects from OAS or installing the operator itself (useful in k8s without olm), manage kuadrant APIkeys (which are secrets with extras), healthchecks...

docs/design/modular_installation.md
eguzki: Authorino and AuthConfig CRDs (together with limitador CRDs) are part of the superset kuadrant API

docs/design/modular_installation.md
eguzki: Yes, the authenticated rate limit is the best example we have so far. 

PD: sorry for the expressiveness, doing my best

eguzki:
Thanks all for the feedback. Clearly, there is room for improvement.

I will go for the second round for this document taking into account all the comments dropped.

docs/design/modular_installation.md
eguzki: ok, removed the "Development" deployment option. That is a property that could be activated in any of the deployment options available.

docs/design/modular_installation.md
eguzki: After re-reading this, I guess I answered too quickly thinking about kuadrant configuration objects (APiProduct, APIs, AuthConfigs, RateLimits)... but maybe we can add some first draft for the `kuadrant CRD`.  @mikenairn do you have already something in mind? 

Something very basic would be:

```yaml
apiVersion: install.kuadrant.io/v1alpha1
kind: Kuadrant
metadata:
  namespace: kuadrant-system
  name: kuadrantcontrolplane
spec:
  profile: full
  components:
     authorino: {}
     limitador: {}
```
I wonder if we need to add some customization of istio. Latest discussions held in other channels indicated that istio would not be owned by kuadrant and should be a given in the cluster, so customization of istio does not make sense. Adding a custom and additional istio gateway would be an option

maleck13: This is an  interesting point of discussion. From a getting started point of view, it may make sense to have the option to install istio for now I think lets point ppl at the istio docs to do  that. As you say taking a gateway name and namespace or perhaps label selector would be needed. Then we either create it if doesn't exist use it if it does exist (for current impl we create our envoyfilters etc and apply to this gateway workload).

maleck13 (approved):

PR #2, opened 2021-12-21
Kuadrant api design v2

eguzki:


docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+* Rate limit is enforced to all services and endpoints defined in a APIProduct. It cannot be specified route/path wise.
+* Only high level rate limit directives are defined (`global`, `perRemoteIP` and `authenticated`). There is no option to specify with a fine-grained rate limit API.
+* Authentication is enforced to all services and endpoints defined in a APIProduct. It cannot be specified route/path wise.
+* Authentication is limited to `API key` and `OpenID Connect`.
+* Authorization is not available.
+
+## High Level Goals
+
+* Expose simple API, API management oriented, to protect customer owned API's.
```
maleck13, line 25: Perhaps add support generating as much boiler plate as possible from a provided API Spec?

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+* Simple API can be extended with a fine-grained API to leverage full featured Authorino API and Limitador API.
+* Path/Operation based protection (i.e. authentication and rate limit)
+
+## Proposed Solution
```
maleck13, line 29: I wonder would it help to capture a set of use cases that the API will help solve?

``` As an API Provider, I want to be able to apply a set of authenticated rate limiting plans based use specific attributes (for example a JWT claim ) so that I can allow different tiers of access to my API```

``` As an API Provider I want to be able to apply both a set of authenticated and unauthenticated rate limit plans so that I can have a "try for free" tier that doesn't require the user to sign up and so that I can have unauthenticated APIs freely available but protected ```

File: docs/design/kuadrant-api-v2.md [hunk 3]
```diff
@@ -0,0 +1,178 @@
...
+
+### API Product
+
+**Custom Resource Definition GVK**
+
+Currently, as of kuadrant v0.2.0, APIProduct [Group Version Kind](https://book.kubebuilder.io/cronjob-tutorial/gvks.html)
+is: [apiproducts.v1beta1.networking.kuadrant.io](https://github.com/Kuadrant/kuadrant-controller/blob/v0.2.0/doc/architecture.md#api-product-crd)
+
+Taking into account kuadrant is not providing networking level API, but **API Management** level API,
+the proposed GVK is: **apiproducts.v1alpha1.apim.kuadrant.io**.
```
maleck13, line 45: +1

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+**Domains**
+
+APIProduct resource will be configured with a list of destination domains or hostnames to which
+traffic is being sent.
+
+It is the responsability of cluster operators and API owners to forward *îngress* traffic for those
```
maleck13, line 52: +1 we shouldn't be creating the route IMO the application developer should define that

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+specific domains to the kuadrant gateway. A kuadrant gateway would be a gateway configured by the
+kuadrant control plane, having as input the API Product resource.
```
maleck13, line 54: This could be an existing gateway right?

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+For example, cluster operators can deploy required (cloud provider level) networking resources to
```
maleck13, line 56: right so it is up to cluster admins to decide on how traffic will hit the gateway. It could be via a route->haproxy->gateway->Kuadrant pieces
More specifically, each APIProduct will tell us which envoy based gateway it is using Kuadrant will then configure the envoy behind that gateway to ensure your API is protected.

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+Each *API Product* can hold multiple OAS docs references. The *API Product* `spec` field includes references
+in the form of URLs and config map references. The actual content may be included in the `status` field.
+
+Kuadrant will extend the OAS doc leveraging
+[OAS custom extensions](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#specificationExtensions)
+`x-kuadrant-*`, initially for rate limit purposes. In the future, OAS extensions could be leveraged
+for other API management features, for instance usage plans. More about the kuadrant custom
+extensions in the spec reference section.
+
+In case of lacking one OpenAPI doc, kuadrant will provide tooling for easy generation of OpenAPI docs.
```
maleck13, line 78: Should we be doing this? I think we have to end users. 1) I know all about OAS and use them 2) I don't want or care about OpenAPI specs but know all about k8s and want to use the resources

maleck13: I feel we should support both

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+**Backends - Upstreams**
+
+The protected API source, owned by the kuadrant customer, can be specified in two forms:
+* A (base) URL
+* A kubernetes service reference (name and namespace).
+
+When the source is a kubernetes service, kuadrant provides annotations and labels that can be added
+to the service. These annotations and labels can specify, for example, the OAS reference.
+These annotations and labels aim to make the `spec` of the *API Product* resource simpler to specify.
```
maleck13, line 90: to encourage committing resources to version control. What if the controller didn't generate the objects in the cluster but instead we had a CLI that generated the resources this could then be used as part of CI/CD flow and also reviewed in code review etc.
In this scenario the controllers job is to just bring the cluster into the state declared by the resources

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+two steps involve kuadrant, but all three of them are required to have the system working.
+For completeness, all of them are listed here.
+
+1. Operation based authentication config
+
+The customer needs to specify which API operation (method + path) requires authentication and
+which type of authentication (API key, oauth2, oidc, http basic, etc). OpenAPI spec fully covers
+this use case and kuadrant can read this configuration from OAS doc. For any authN type not covered
+by OAS 3.X, kuadrant could provide extensions for added authentication types. OpenAPI 3.x does
+not support authentication requirements at path level, only global or operation level.
```
maleck13, line 104: I found this paragraph a bit unclear. Initially seemed OAS supported a path based authentication but then the last sentence seems to state it doesnt?

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+## Problem Statement
+
+Currently, the [latest](https://github.com/Kuadrant/kuadrant-controller/releases/tag/v0.2.0) release of
+kuadrant API is based on [apiproducts.v1beta1.networking.kuadrant.io](https://github.com/Kuadrant/kuadrant-controller/blob/v0.2.0/doc/architecture.md#api-product-crd)
+and [apis.v1beta1.networking.kuadrant.io](https://github.com/Kuadrant/kuadrant-controller/blob/v0.2.0/doc/architecture.md#api-crd)
+custom kubernetes resource APIs. Together with these CRDs, there is a bunch of kuadrant service
+annotations and labels available in order to provide the kuadrant
+[service discovery](https://github.com/Kuadrant/kuadrant-controller/blob/v0.2.0/doc/service-discovery.md)
+feature. All together, kuadrant provides routing, basic and global rate limiting and, finally,
+global authN (not authZ) features to protect customer owned APIs. While, good enough for a wide set
```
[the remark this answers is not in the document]
eguzki: I mean the current (v0.2.0) kuadrant implementation and API does not provide authZ. in v0.2.0, the customer cannot customize authconfig resources. Those resources are created and owned by the kuadrant controller. The new API design, on the other hand, via templates the customer can add authZ policies

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+**RateLimit Template**
+
+The configuration of the API rate limit requires *two* steps.
+
+1. Path/Operation based rate limit config
+
+The customer needs to define which API paths/operations require rate limit enforcing.
+Furthermore, the customer needs to define what type of rate limit needs to be enforced at
+path/operation level. Global rate limit is a particular case where all API operations require
+enforcing rate limit of a particular rate limit type.
```
maleck13, line 140: for rate limiting I think we need to allow for headers also

docs/design/kuadrant-api-v2.md
eguzki: good contribution :+1:

docs/design/kuadrant-api-v2.md
eguzki: Sure. The new API needs a mechanism to link a given APIProduct with one (or multiple??) gateway(s)

docs/design/kuadrant-api-v2.md
eguzki: exactly, that is what I wanted to express.

docs/design/kuadrant-api-v2.md
eguzki: For the second use case, currently kuadrant supported the "PathMatch" option. If the user does not have OAS, it will need to provide some info:
* path match (`GET /toystore/*`) or set of operations (` [ "GET /toystore/dolls", "GET /toystore/egames" ] `)
* K8s service or upstream URL
* Desired authentication
* Desired rate limit 

With that information, my proposal was to have some tooling available to generate OAS. Can be the CLI, online web app, custom resource for that... With the OAS in place, the user can protect with kuadrant.

docs/design/kuadrant-api-v2.md
eguzki: I see your point and definitely something to have in mind. 

Asking the customer to review istio/envoy resources is asking the customer to have knowledge at the networking layer. I do not think it makes sense, but we can discuss further. 

From my current perspective (changing every week or so), kuadrant is a bridge between API management layer and networking layer. Understands the API owner's language (OpenAPI) and configures the networking layer to make it happen. So, asking the customer to review the networking layer manifests breaks that model. It's like telling the customer: "here you have istio, Authorino and RateLimit, go and configure yourself for your own needs".

docs/design/kuadrant-api-v2.md
eguzki: OpenAPI 3.x supports operation level auth requirements. however, do not ask me why, you cannot say "I want to apply this auth requirement to `/toystore/dolls` for all HTTP methods". Instead you need to duplicate the auth requirement in all HTTP methods for that path.

I will rewrite the paragraph

docs/design/kuadrant-api-v2.md
eguzki: It could be. But then, OAS cannot be used to specify *where* to apply rate limit. Instead, kuadrant needs to implement a rule matching system for rate limit. It would be similar to the [Rules of istio's Authorization Policy](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Rule)

For the first implementation, I would stick to OAS. But definitely, that can be a real user requirement.

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+## High Level Goals
+
+* Expose simple API, API management oriented, to protect customer owned API's.
+* Simple API can be extended with a fine-grained API to leverage full featured Authorino API and Limitador API.
+* Path/Operation based protection (i.e. authentication and rate limit)
+
+## Proposed Solution
+
+The kuadrant API will be oriented to meet API management requirements. In that context, the kuadrant
+API will be designed around the OpenAPI standard.
```
guicassolato, line 32: I wonder whether OpenAPI is one of many to support in the future or it can indeed drive the direction the Kuadrant API is designed.

Other types of API specs to mention: (Google) protocol buffers for (g)RPC, AsyncAPI, GraphQL, WSDL, Kafka Connect Schema. Will we ever be able to conciliate all this into one API or are we talking here about other "Kuadrants"?

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+It is the responsability of cluster operators and API owners to forward *îngress* traffic for those
+specific domains to the kuadrant gateway. A kuadrant gateway would be a gateway configured by the
+kuadrant control plane, having as input the API Product resource.
+
+For example, cluster operators can deploy required (cloud provider level) networking resources to
+forward traffic to the cluster ingress gateways. Then, the API owners can create (cluster level)
+networking resources, like kubernetes Ingress or Gateway API HttpRoutes, to route that traffic to
+the kuadrant gateways. In this example only *North/South* traffic was referenced. By design,ç
+it could also be *East/west* traffic. By design, there is nothing preventing from having *East/West*
+traffic being protected by kuadrant. This is mainly because kuadrant is not a networking API.
```
guicassolato, line 61: +1

docs/design/kuadrant-api-v2.md
guicassolato: The link between a path and a security scheme (e.g. authn) can be through an operation and its "security requirements":
- pathA → operation1 → securityRequirement → securityScheme_x
- pathB → operation1 → securityRequirement → securityScheme_x
- pathC → operation2 → securityRequirement → securityScheme_x
- pathD → operation3 → securityRequirement → securityScheme_y

In the above, pathA, pathB and pathC all share the same authentication scheme, whereas pathD can have its own.

docs/design/kuadrant-api-v2.md
```diff
@@ -0,0 +1,178 @@
...
+
+The *API Product* resource allows to define multiple AuthConfig templates.
+
+To illustrate this with examples, the API Key auth type would require `AuthConfig` template with
+`spec.identity[].apiKey` object. The OIDC auth type would require `AuthConfig` template with
+`spec.identity[].oidc` object.
+
+Kuadrant can add some validations to prevent inconcistencies. For instance, if the OAS doc specifies
+that API key is being required for some API operations, then kuadrant can make sure that the
+`AuthConfig` template includes `apiKey` identity type.
```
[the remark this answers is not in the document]
guicassolato: https://github.com/Kuadrant/authorino/pull/206

PR #3, opened 2022-09-20
Initial RFC process' RFC

alexsnaps:
# Summary

This is an attempt at streamlining and formalizing the addition of features to Kuadrant and its components: `Authorino`, `Limitador` and possible more to come. It describes a process, Request For Comment (RFC). That process would be followed by contributors to Kuadrant in order to add features to the platform. The process aims at enabling the teams to deliver better defined software with a better experience for our end-users.

# Motivation

As I went through the process of redefining the syntax for `Condition`s in `Limitador`, I found it hard to seed people's mind with the problem's space as I perceived it. I started by asking questions on the issues itself that didn't get the traction I had hoped for until the PR was eventually opened.
This process should help the author to consider the proposed change in its entirety: the change, its pros & cons, its documentation and the error cases. Making it easier for reviewers to understand the impact of the change being considered.
Further more, this keeps a "written" record of "a decision log" of how a feature came to be. It would help the ones among us who tend to forget about things, but would be of incommensurate value for future contributors wanting to either understand a feature deeply or build upon certain features to enable a new one.

# Guide-level explanation

A contributor would start by following the template for a new _Request For Comment_ (RFC). Eventually opening a pull request with the proposed change explained. At which point it automatically becomes a point of discussion for the next upcoming _technical discussion_ weekly call. 
Anyone is free to add ideas, raise issues, point out possible missing bits in the proposal before the call on the PR itself. The outcome of the _technical discussion_ call is recorded on the PR as well, for future reference. 
Once the author feels the proposal is in a good shape and has addressed the comments provided by the team and community, they can label the RFC as _FCP_, entering the Final Comment Period_. From that point on, there is another week left for commenters to express any remaining concerns. After which, the RFC is merged and going into _active_ status, ready for implementation.

# Reference-level explanation

See the PR's content

# Drawbacks

The process proposed here adds overhead to addition of new features onto our stack. It will require more upfront specification work. It may require doing a few proof of concepts along the initial authoring, to enable the author to better understand the problem space.

# Rationale and alternatives

What we've done until now, `investigation`s have been less formal, but I'm unsure how much of their value got properly and entirely captured. By formalizing the process and having a clear outcome: a implementable piece of documentation, that address all aspects of the user's experience look like a better result. 

# Prior art

The entire idea isn't new. This very proposal is based on prior art by [`rust-lang`](https://github.com/rust-lang/rfcs) and [`pony-lang`](https://github.com/ponylang/rfcs). This process isn't perfect, but has been proven over and over again to work.

# Unresolved questions

- A week for the _FCP_ seems a lot and very little at the same time… should we revisit this?
- Having two core team member accepting an RFC is… acceptable? Should it be more? Less?
- ~Should this all go under `Kuadrant/rfcs`?~ `Kuadrant/architecture` it is…
- What does it mean for [`kcp-glbc`](https://github.com/Kuadrant/kcp-glbc) ?

# Future possibilities

I certainly see this process itself evolving overtime. I like to think that this process can itself be supporting its future changes…

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -1,0 +1,81 @@
+# RFC Template
+
+- Feature Name: (fill me in with a unique ident, `new_feature`)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Explain the proposal as if it was implemented and you were teaching it to Kuadrant user. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how a user should *think* about the feature, and how it would impact the way they already use Kuadrant. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing and new Kuadrant users.
+
+# Reference-level explanation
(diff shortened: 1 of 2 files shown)


maleck13 (approved):
I like the idea of reviewing outstanding RFCs during the technical/community call. That way we can keep any that are waiting in peoples minds

PR #4, opened 2022-11-10
Single merged policy

alexsnaps:


rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,228 @@
...
+types.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- How do we deal with multiple `Gateway` having different set of policies attached to them
+  (semantically or atomically?)?
+  - Can we inject the proper config in each of them? Only if all is applicable? What if they are not?
+    What when that changes over time?
+  - Can these then be backed by a single e.g. `Limitador`? (in that case I think yes, as we could
```
eguzki, line 217: I would say yes, multiple gateways can be backend by a single Limitador. Currently, the `namespace` field has a prefix withthe name/NS of the gateway, so limits are well scoped to each gateway. This an implementation detail that can be discussed, of course.

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,228 @@
...
+but they are an implementor and as such can extend the
+[`HTTPRouteFilter`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter)
+types.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- How do we deal with multiple `Gateway` having different set of policies attached to them
+  (semantically or atomically?)?
+  - Can we inject the proper config in each of them? Only if all is applicable? What if they are not?
```
eguzki, line 215: I do not follow this. 

Yes, we can inject proper config to each gateway. So, 
RLP A -> Gateway A

RLP B -> Gateway B

eguzki (commented):
Overall I like the proposal and I am willing to implement that. 

However, I see a potantial caveat on this approach regarding RLP's targeting a gateway. As a cluster administrator I have detected unusual traffic for `suspect-subdomain.domain.com` a I want to rate lmit that traffic for "Host: suspect-subdomain.domain.com". Assuming there is a HTTPRoute for this tenant for `*.domain.com`

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,228 @@
...
+`HTTPRoute` all applicable policies of a `kind` have been merged into a single one _and_ that single
+policy then applies to the whole traffic of that route.
+
+The current approach, which lets each policy define an arbitrary subset of the traffic to which it
+applies, makes it impossible to know how to merge the different policies into one. Each policy's set
+requires to know how to applicable it becomes to the union and/or the intersection of the said
+subsets. While a heuristic could be applied, there are use-cases where the other option will be
+desirable. The question then becomes which policy is the authority as to decide which it should be. In
+order to avoid having to answer these questions, this proposal looks at not letting a policy
+discriminate which part of traffic it applies to: it's the whole traffic, always.
```
maleck13, line 37: Ok so in this case it would be up to me to define the auth or rate limits based on the path/method as part of defining the limits within the policy or the authschema? So if I wanted no limits on a particular endpoint or no auth on an endpoint

maleck13: ok reading on further I see that is it is a much more limited API. I can see how this will help with overrides and defaults, but my concern is that it is too limited to really be useful?

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,228 @@
...
+[`HTTPRouteRule`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteRule)
+define on that `HTTPRoute`. All the traffic matched by _any_ `HTTPRouteRule` of a given `HTTPRoute`
+will be affected by which ever `RateLimitPolicy` or `AuthPolicy` you attach to it.
+
+Any Kuadrant `Policy` will apply to the entirety of the traffic that is routed through the `HTTPRoute`
+it is eventually attached to. A `Policy` can be attached directly to an `HTTPRoute`. It will apply its
+specification according to the `override` and `default` section of the `Policy`'s `spec`. Let's use a
+simple example:
+
+```yaml
```
maleck13, line 59: so I like the simplicity being aimed at here but I worry we will need to allow some form of conditional here
example if we were to always send common descriptors (if we agree they are common) e.g Headers. Would we see a way for someone to define 

```
limits:
        - conditions:
            - "req.path=='/checkout'"
            - "req.method=='POST'"
          maxValue: 5
          seconds: 10
          variables: []
```

maleck13: But I feel you are aiming at something very different. I had thought perhaps by applying a policy to all traffic, it had meant all traffic would be sent to the implementing service and then the implementing service could make a conditional decision

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+>**Note:** The status block of a policy should reflect the merged state of a policy as per other
+>policies at that level of the hierarchy.
+
+Any other `HTTPRoute` attached to the `my-gw` would still use the 42 rps limit. If we now delete the
+`FakeRLPolicy` attached to `my-gw`, the resulting policy would look like this:
+
+```yaml
+enabled: true
```
guicassolato, line 158:
```suggestion
enabled: false
```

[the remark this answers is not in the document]
alexsnaps: No, it is set as `true` as an override on the GW attached policy. That's what I describe in the paragraph above...

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,285 @@
...
+Attachment hierarchy. Targeting a complete `HTTPRoute` is the equivalent of targeting each indidividual `HTTPRouteRule`s
+that composes the route.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+So there are a couple of things to consider here:
+
+- status reporting on "not yet" existing `HTTPRouteRule`, e.g. a Policy declares a set of `matches` that don't (yet?)
+  exist. The policy must _not_ be enforced... yet. But could eventually. This needs proper reporting.
```
didierofrivia, line 235: An alternative could be report the policy as `invalid`, that could simplify things at least at the beginning... a typical error would be `RateLimitPolicy/AuthPolicy rules don't match the HTTPRouteRule definition`

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,285 @@
...
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how
+the detailed proposal makes those examples work.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- The use cases this approach doesn't address… 
+- Not very DRY... the need to keep `HTTPRouteRule`s in sync between the Policy and the `HTTPRoute` is suboptimal.
```
didierofrivia, line 251: This is probably the most discouraging point, but the good thing is we could mitigate a bit with status reporting and ux tooling.

guicassolato:
I'd like to propose an exercise to put this to proof.

First, of course, we'd need to agree on the exact set of use cases to cover. IMO, the following one seems "real" enough. At the same time, it's simple (with only 2 Gateways, 2 HTTPRoutes, not too many matching rules) and yet possibly intricate enough.

### Exercise

Given the following initial network resources:

```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gateway-1
  namespace: gwns
spec:
  gatewayClassName: internal
  listeners:
    - hostname: *.acme.internal
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gateway-2
  namespace: gwns
spec:
  gatewayClassName: internet
  listeners:
    - hostname: *.acme.com
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: toys
  namespace: toys
spec:
  parentRefs:
    - kind: Gateway
      namespace: gwns
      name: gateway-1
    - kind: Gateway
      namespace: gwns
      name: gateway-2
  hostnames:
    - toys.acme.com
    - toys.acme.internal
  rules:
    - backendRefs:
        - kind: Service
          name: toys
    - matches:
        - headers:
            - name: X-Env
              value: canary
      backendRefs:
        - kind: Service
          name: toys-canary
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: telemetry
  namespace: telemetry
spec:
  parentRefs:
    - kind: Gateway
      namespace: gwns
      name: gateway-1
  hostnames:
    - *.telemetry.acme.internal
  rules:
    - matches:
        - method: GET
        - method: POST
      backendRefs:
        - kind: Service
          name: telemetry
```

What are the set of policies (RLP and KAP) and possibly required changes to the network resources to implement the following use cases?

1. RLP _override_ **500 rps** on all routes matching `*.acme.com`.
2. RLP _default_ **100 rps** on **POST** requests to all routes matching `*.acme.com`.
3. KAP _override_ **IP deny-list** on all requests to `*.acme.com`.
4. KAP _default_ **X.509 certificate authentication** on all requests to `*.acme.internal`.
5. RLP _default_ **150 rps** on all requests to `toys.acme.(com|internal)`.
6. RLP _default_ **50 rps** on all requests to `toys.acme.(com|internal)` containing `X-Env: canary` request header.
7. RLP _override_ **unlimited rps** on all requests to `toys.acme.internal/admin/*`.
8. KAP _override_ **API key authentication** on all requests to `toys.acme.com`.
9. KAP _override_ **DELETE** requests **forbidden** at `toys.acme.com`.
10. KAP _override_ all requests **forbidden** at `toys.acme.com/admin/*`.
11. KAP _override_ **JSON pattern-matching authorisation** to check path param {org_name} matches value stored in API key or X.509 cert on all requests to `toys.acme.(com|internal)/[^admin/]orgs/{org_name}/*`. 
12. KAP _override_ **K8s SAR authorisation** on all requests to `toys.acme.internal/admin/*`.
13. KAP _default_ **OIDC/JWT authentication** on all endpoints matching `*.telemetry.acme.internal`.
14. KAP _override_ additional **API key authentication** to all requests to `foo.telemetry.acme.internal`.

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,285 @@
...
+    - headers:
+      - type: Exact
+        name: env
+        value: production
+    backendRefs:
+    - name: bar-svc
+      port: 8080
+```
+
+It is important to consider that only _one_ rule will ever be match for any given requests. It makes for a simpler model to have only one "Policy" to also only ever being match, aligning the both behaviors. To apply a policy on the first rule only, the user needs to duplicate that match on the policy targeting `bar-route`:
```
eguzki, line 203: > the user needs to duplicate that match on the policy targeting `bar-route`

This is essentially what we have now, isn't it?

rfcs/0000-policy_target_scope.md
```diff
@@ -0,0 +1,285 @@
...
+    backendRefs:
+    - name: foo
+      port: 8080
+```
+
+With `foo-route` declaring a single hostname (`example.com`) and a single `HTTPRouteMatch` (anything
+starting with `/foo`), the rate limiting will apply to the same traffic: `example.com/foo*`, so that
+`foo` will not see more than 42 rps on port `8080`.
+
+Given the `override` on the `my-gw` `FakeRLPolicy` there no way for our route to disable rate limiting
```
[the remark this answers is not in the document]
alexsnaps: But it would "just" be up to the GW to express this properly, i.e. using a `default` rather than an `override`. Also, there might be a need to say `10` (i.e. less than 42), say when being DDoS'ed, so that'd be an override... The fact that the API lets users express what we think are "none sensible" things, doesn't invalidate it. 

But I think we could also have Kuadrant "merge" the policies based of the most restrictive wins and "only" support `defaults` initially.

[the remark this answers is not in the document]
eguzki: > Also, there might be a need to say 10 (i.e. less than 42), say when being DDoS'ed, so that'd be an override...

That is a valid use case and it is covered by "most restrictive policy wins"

> But I think we could also have Kuadrant "merge" the policies based of the most restrictive wins 

That is what I call "conditional defaults/overrides" which is worth elaborating/investigating more, but I find it as a complex API to expose to the customers. 

> "only" support defaults initially.

Is this a valid use case? I mean, a gateway provider saying I want (default) 10 rps but I do not care if you as API provider sets 1000 rps. My point being, if the gateway provider applies 10 rps, that 10 is not a magical number it is something specific for the existing resources and constraints.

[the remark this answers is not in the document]
maleck13: We probably need more real world use cases to be sure.  A clear use case is a gateway provider/infra provider setting a global limit for the gateway to protect it lets say 1000 RPS and wanting that to be a "override" limit. This is a use case we have in other services where we have used limitador to do this at the gateway level as the infra provider in order to enforce a limit based the tier. 
So then the question and use case comes for a default RL. This is less clear as @eguzki points out. That said, I can imagine as a "gateway admin", I don't want any one service to be able to just accidentally consume this entire limit without clearly expressing this intention. I don't want any endpoint deployed to my gateway that some random client can accidentally get into a loop and swamp the gateway. So I now set a sane but restrictive default based on remote ip or some common property to a low number lets say 50 RPS. Now I know if an endpoint is created without any RLP, it can only at worst consume 50 RPS. 
So now it is up to the service provider/developer to say explicitly "I need more than 50 RPS  I need 200 ~RPS" for my service. So the piece I see as missing here is setting an "upper limit" override. IE by default its 100 but you can override it up to 250 but beyond that we need to have a serious conversation. 
@eguzki Does the above make sense in your world? I don't claim to have all the correct use cases here but the above seems sane to me. Interested in your thoughts on the default and being able to set a "upper limit" override.
I think it is also important that we don't think of RLP as the only control involved. As a gateway admin, depending on the environment, I am going to want to setup monitoring so that I can identify what limits are set, and what traffic is flowing to certain endpoints. This will allow me to identify a potential problem before it happens. Also I might have guidelines in place for the teams using my gateway and perhaps I even have an SRE member reviewing any new limits set as part of a git ops flow etc. I think we need to think of RLP as part of a whole rather than the whole.

alexsnaps:
superseded by https://github.com/Kuadrant/architecture/pull/58

PR #5, opened 2022-11-14
Multiple kuadrant instances

eguzki:


rfcs/0000-wire-policies-with-backend-services.md
```diff
@@ -0,0 +1,181 @@
...
+Until there is a real use case and it is strictly necessary, this scenario is discarded for implementing kuadrant's protection.
+The main reason is about complexity. It is already complex enough to reason about rate limiting and auth services having a single stage. Adding multiple rate limiting stages, or hitting multiple Limitador instances in a single stage (doable with the WASM module) makes it too complex to reason about observed behavior. Currently there is no use case to require that complex scenario.
+
+A kuadrant instance includes:
+
+* One Limitador deployment instance
+* One Authorino deployment instance
+* A list of dedicated gateways. Those gateways cannot be shared between multiple kuadrant instances.
+
+![](https://i.imgur.com/QdeCYs6.png)
```
eguzki, line 80: The diagram can be seen also from the branch: https://github.com/Kuadrant/architecture/blob/wire-policies-with-backing-service/rfcs/0000-wire-policies-with-backend-services.md

rfcs/0000-wire-policies-with-backend-services.md
```diff
@@ -0,0 +1,188 @@
...
+
+Highlights:
+* The Kuadrant instance is not enclosed by k8s namespaces.
+* One gateway can belong to (be managed by) one and only one kuadrant instance.
+* One kuadrant instance does not own rate limit policies or auth policies.
+* Each kuadrant instance owns one instance (possibly multiple replicas, though) of Limitador and one instance of Authorino. Those instances are shared among all gateways managed by the kuadrant instance.
+* The traffic routed by HTTPRoute 1 through the gateway A will be protected by RLP 1 and KAP 1, using Limitador and Authorino instances located at the namespace K1.
+* The traffic routed by HTTPRoute 2 through the gateway B will be protected by RLP 2 and KAP 2, using Limitador and Authorino instances located at the namespace K1.
+* The traffic routed by HTTPRoute 2 through the gateway C will be protected by RLP 2 and KAP 2, using Limitador and Authorino instances located at the namespace K2.
+* The HTTPRoute 2 example shows that when the traffic for the same service is routed through multiple gateways, at least for rate limiting, Kuadrant cannot keep consistent counters. The user would expect X rps and actually it would be X rps per gateway.
```
guicassolato, line 92: IMO, this is the hardest limitation to sell? One who knows about the [global RL protocol](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting) may find counterintuitive that 2 gateways cannot share the same Limitador instance.

Could synchronization of the counters across multiple instances of Limitador mitigate this (as it makes fairly transparent that for 2 GWs there are actually 2 instances of Limitador)? Or perhaps it's better is the design allowed for a Kuadrant instance to take control of more than one GW (which seems possible with the implementation proposed based on label selectors)?

rfcs/0000-wire-policies-with-backend-services.md
```diff
@@ -0,0 +1,188 @@
...
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+  name: kuadrant-sample
+spec:
+  controlPlane:
+    image: quay.io/kuadrant/kuadrant-operator:mytag
+  limitador:
+    image: quay.io/kuadrant/limitador:mytag
+  authorino:
```
guicassolato, line 122: This can be typed as Authorino Operator's [`AuthorinoSpec`](https://github.com/Kuadrant/authorino-operator/blob/4996d89fd18d7d1588e8aeb090a7bcb4968b5fb9/api/v1beta1/authorino_types.go#L119-L136).

rfcs/0000-wire-policies-with-backend-services.md
eguzki: I think it does not apply anymore. Check the new content.

rfcs/0000-wire-policies-with-backend-services.md
eguzki: this was out of scope, so removed

rfcs/0000-wire-policies-with-backend-services.md
```diff
@@ -0,0 +1,188 @@
...
+
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+  name: kuadrant-sample
+spec:
+  controlPlane:
+    image: quay.io/kuadrant/kuadrant-operator:mytag
+  limitador:
```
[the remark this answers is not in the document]
eguzki: out of scope, so removed

eguzki:
looking for review @Kuadrant/engineering

rfcs/0000-multiple-kuadrant-instances.md
```diff
@@ -0,0 +1,152 @@
...
+Technically, the Kuadrant policies do not belong to any Kuadrant instance. At any moment of time, one policy can switch the targeted network resource specified in the `spec` from one gateway to another. Directly or indirectly via the HTTPRoute. The target references are dynamic by nature, so is the list of gateways to which kuadrant policies should apply.
+Thus, the Kuadrant's control plane needs a procedure to associate a policy with **one** kuadrant instance at any time. When the control plane knows which kuadrant instance is affected, the policy rules can be used to configure the Limitador and Authorino instances belonging to that kuadrant's instance. Since the associated kuadrant instance of a policy is dynamic by nature, this procedure must be executed on  every event related to the policy.
+
+When the policy's `targetRef` targets a Gateway, there is a direct reference to the gateway.
+
+When the policy's `targetRef` targets an HTTPRoute, Kuadrant will follow the [`parentRef`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.CommonRouteSpec) attribute which should be a direct reference to the gateway or gateways.
```
didierofrivia, line 92: This means that indirectly, a KAP or RLP, would be applied to more than 1 Kuadrant instance. As mentioned before, the control plane in charge of associating policy with 1 Kuadrant instance, will need to decide which policy apply in case there's another HTTPRoute sharing one or more of `parentRef` gateways

didierofrivia: OK, it's mentioned that this use case is [not supported](https://github.com/Kuadrant/architecture/pull/5/files#diff-416c18ad603a7ea22bfb07cf5c9434220f70097b4746c60e4bb1b527496467bdR51) , still we might need to provide a way to mitigate since one might not have access to modify the network topology

rfcs/0000-multiple-kuadrant-instances.md
```diff
@@ -0,0 +1,152 @@
...
+# Motivation
+[motivation]: #motivation
+
+The main benefit of multiple Kuadrant instances in a single cluster is that it allows dedicated Kuadrant's services for tenants.
+
+Dedicated Kuadrant deployment brings lots of benefits. Just to name a few:
+* Protection against external traffic load spikes. Other tenant's traffic spikes does not affect Authorino/Limitador throughput and delay as it would when shared.
+* No need to have cluster administrator role to deploy a kuadrant instance. One tenant administrator can manage gateways, Limitador and Authorino instances (including deployment modes).
+* The cluster administrator gets control and visibility across all the Kuadrant instances, while the tenant administrator only gets control over their specific gateway(s), Limitador and Authorino instances.
+* (looking for ideas for more benefits)...
```
maleck13, line 24: I think the key benefit is protection against noisy neighbour and isolation particularly in the auth context.

rfcs/0000-multiple-kuadrant-instances.md
```diff
@@ -0,0 +1,152 @@
...
+
+When the policy's `targetRef` targets a Gateway, there is a direct reference to the gateway.
+
+When the policy's `targetRef` targets an HTTPRoute, Kuadrant will follow the [`parentRef`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.CommonRouteSpec) attribute which should be a direct reference to the gateway or gateways.
+
+Given a gateway, Kuadrant needs to find out which Kuadrant's instance is managing that specific gateway. By design, Kuadrant knows it is only one. There are at least two options to implement that mapping:
+* Read all Kuadrant CR objects and the first one that matches label selector.
+  * This approach works as long as the control plane ensures that each gateway is matched by only one kuadrant gateway selector. The control plane must reject any new kuadrant instance matching a gateway already "taken" by other kuadrant instance.
+* Add annotation in the gateway with a value of the Name/Namespace of the Kuadrant CR.
+  * This approach is commonly used. Requires annotation management.
```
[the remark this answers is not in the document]
didierofrivia: We could define the list of gateways that meant to be managed in the Kuadrant CR.

rfcs/0000-multiple-kuadrant-instances.md
```diff
@@ -0,0 +1,152 @@
...
+
+![](https://i.imgur.com/yChVsT6.png)
+
+### All the gateways referenced by a single policy must belong to the same kuadrant instance
+
+The Gateway API allows, in its latest version  [v1beta1](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.CommonRouteSpec), an HTTPRoute to have multiple gateway parents. Thus, a kuadrant policy might technically target multiple gateways managed by multiple kuadrant instances. Kuadrant does **not** support this use case.
+
+![](https://i.imgur.com/ZpsBf4i.png)
+
+The main reason is related to the rate limiting capability. The limits specified in the RateLimit Policy would be enforced per kuadrant instance basis (provided by Limitador instance). Thus, traffic hitting one gateway would see different rate limiting counters than traffic hitting the other gateway. The user would expect X rps and actually it would be X rps per gateway. For consistency reasons, when this configuration happens, the control plane will reject the policy.
```
alexsnaps, line 55: While I think this is a good first step and we indeed should reject such policies initially, I think there might be a way to support this as we get to architect for multiple cluster support. The use-case isn't much different. What I'm thinking is that each limitador would know about each other. Upon hitting a "shared" limit, they'd start sharing the counters for it. Which, other than latency being lower and cross communication being easier, there isn't much differences between the two applications. 

Sharing this thought here as an FYI and sharing a possibly path forward. Again, it looks perfectly acceptable to have that limitation initially.

rfcs/0000-multiple-kuadrant-instances.md
```diff
@@ -0,0 +1,152 @@
...
+- What is the impact of not doing this?
+
+This design is a step forward in a consistent API to protect other service APIs. It makes easier to protect any API, no matter traffic nature. Either north-south or east-west. It makes easier to have the scenario where cluster users deploy their own (not ingress) gateways and enable API protection declaratively.
+
+# Prior art
+[prior-art]: #prior-art
+
+TODO
+
+# Unresolved questions
```
alexsnaps, line 131: It might already be work in progress, but a specification of how errors (and which error cases) would be surfaced to the user would be good. For instance, a policy is applied to a `HTTPRoute` alright. The route gets an additional `Gateway` wired to it… what's the behavior? How does the user know?

guicassolato: > a policy is applied to a `HTTPRoute` alright. The route gets an additional `Gateway` wired to it… what's the behavior? How does the user know?

Currently, the addition of a new `Gateway` to a `HTTPRoute` targeted by a policy is not reflected in the policy status, just like any other event of modifying the `HTTPRoute` that is handled successfully by the controller. The controller basically only reports _"HTTPRoute is protected"_ or reconciliation errors. But even with the error messages, they are not always helpful to users.

I think this is already important today and even more so in a context of multiple Kuadrant instances per cluster. Users could benefit from the info of which Gateways have been configured for a policy, which were expected to but have not (and why not), in the policy status. Currently, users can only partially work that out by reading in the annotations of the Gateways which policies affect the Gateway.

rfcs/0000-multiple-kuadrant-instances.md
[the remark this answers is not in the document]
guicassolato: I'd suggest a list of `GatewayClasses` instead of `Gateways`.

Users can always "split" one `GatewayClass` into two or more, handled by the same controller or not, to highlight the fact that different configurations apply to each class. In fact, this is encouraged by [GW-API](https://gateway-api.sigs.k8s.io/api-types/gatewayclass/):

> We expect that one or more GatewayClasses will be created by the infrastructure provider for the user. It allows decoupling of which mechanism (e.g. controller) implements the Gateways from the user. For instance, an infrastructure provider may create two GatewayClasses named internet and private to reflect Gateways that define Internet-facing vs private, internal applications.

 Some discussion about this also here: https://github.com/Kuadrant/architecture/pull/7#issuecomment-1397383619

rfcs/0000-multiple-kuadrant-instances.md
```diff
@@ -0,0 +1,152 @@
...
+
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+  name: kuadrant-a
+spec:
+  gatewaysSelector:
+    matchLabels:
+      app: kuadrant-a
```
guicassolato, line 80: I've been thinking about the label selector approach for this application lately. Although it is consistent with other usages such as how Authorino selects AuthConfigs, how AuthConfigs select API key and X.509 cert issuer Secrets, how Istio selects workloads, among others, I think for this particular case of selecting (assigning) Gateways for a Kuadrant instance nevertheless an approach based on the GatewayClasses could be a better fit. A few advantages I see:
- More assertive as the Kuadrant CR would list GatewayClasses by name explicitly;
- Rational topology with individual GatewayClasses sometimes split into two or more, making it clear which groups of Gateways are configured by Kuadrant and which are not;
- Easier to validate that the Gateways assigned to a Kuadrant instance are implemented by a compatible gateway provider (Istio, OSSM, EG, etc), with less (or zero) variations of providers within each group of gateways assigned to a Kuadrant instance. (See also https://github.com/Kuadrant/architecture/pull/7)

rfcs/0000-multiple-kuadrant-instances.md
guicassolato: https://github.com/Kuadrant/kuadrant-operator/issues/140

rfcs/0000-multiple-kuadrant-instances.md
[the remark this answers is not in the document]
eguzki: Kuadrant manages gateway instances, not gateway classes. Kuadrant instance A may manage gateway X and instance B may mananage instance Y. Both X and Y gateways may be share the same gateway class.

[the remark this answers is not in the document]
guicassolato: I guess what I'm saying is that Kuadrant can use the GatewayClass to know which Gateways to manage.

In your provided example, 

> Kuadrant instance A may manage gateway X and instance B may mananage instance Y

1. You can have both instances of Kuadrant listing the one and only GatewayClass; or
2. You can have 2 GatewayClasses, one for each gateway, one for each instance of Kuadrant.

eguzki:
out of scope as of today. Maybe some other day

guicassolato:
> out of scope as of today. Maybe some other day

I would like to reopen this please and clarify first. What is out of scope? The mechanism to filter specific gateways to be in the scope of a Kuadrant instance or supporting multiple gateway providers in the same cluster? I believe the latter should still be in consideration.

eguzki:
This is issue is about kuadrant control plane supporting multiple Kuadrant CR's and the implications of that (related to map gateways to kuadrant instances). `supporting multiple gateway providers in the same cluster` is something else.

guicassolato:
> This is issue is about kuadrant control plane supporting multiple Kuadrant CR's and the implications of that (related to map gateways to kuadrant instances). `supporting multiple gateway providers in the same cluster` is something else.

Roger that, @eguzki. Thanks for explaining. Closing it back then.

PR #7, opened 2022-12-13
Multiple gateway providers in the same cluster

guicassolato:


eguzki:
For me the gateway provider should be per gateway basis instead of per kuadrant instance basis. The gateway, represented by the Gateway API's Gateway resource, should specify the provider. Actually, the `gatewayClassName` was intended for that. I think that the kuadrant control plane should rely on attributes read from the gateway to know how to configure it. If the `gatewayClassName` is not enough, maybe kuadrant can define some label/annotation that must be in the gateway object to have kuadrant configure it for rate limiting and auth.

For me this is the most flexible way. Less prone to errors like trying to configure gateways with some provider specified in the Kuadrant CR and the actual provider is other one.

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+            service: <qualified-name-of-the-authorino-service-for-this-kuadrant-instance>
+            port: <port-number-of-the-grpc-authorino-service-for-this-kuadrant-instance>
+  ...
+```
+
+Moreover, the Operator would as well make sure a [`ServiceMeshMemberRoll`](https://docs.openshift.com/container-platform/4.11/service_mesh/v2x/ossm-create-mesh.html#ossm-member-roll-modify_ossm-create-mesh) resource named `default` exists and includes the namespace of the `Kuadrant` CR in its `spec.members` field. I.e.
+
+```yaml
+apiGroup: maistra.io/v1
+kind: ServiceMeshControlPlane
```
guicassolato, line 113:
```suggestion
kind: ServiceMeshMemberRoll
```

guicassolato:
@eguzki 

> For me the gateway provider should be per gateway basis instead of per kuadrant instance basis. The gateway, represented by the Gateway API's Gateway resource, should specify the provider. Actually, the `gatewayClassName` was intended for that. I think that the kuadrant control plane should rely on attributes read from the gateway to know how to configure it.

The [`gatewayClassName`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway) is just the name of a `GatewayClass` resource, so we'd need to read this other resource. Unfortunately the only thing that the `GatewayClass` resource guarantees is another field called [`controllerName`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayClass), which may or may not gives enough hint to infer the provider. 

> If the `gatewayClassName` is not enough, maybe kuadrant can define some label/annotation that must be in the gateway object to have kuadrant configure it for rate limiting and auth.

An annotation containing the info we need to infer the provider is a valid approach; likely that would be in the `GatewayClass` instead of each `Gateway`, since all gateways under a same `GatewayClass` must be implemented by the same provider.

[This actually gives me another idea...]

> For me this is the most flexible way. Less prone to errors like trying to configure gateways with some provider specified in the `Kuadrant` CR and the actual provider is other one.

You raise a good point here. By just dumping a list of gateway providers in the Kuadrant CR (as I originally proposed), we get nothing that tells us which gateways are provided by each provider. It may suffice to inject configurations that apply to the entire provider, but not enough to cover occasional differences between providers when configuring the gateways themselves.

[...and this is the idea:]

So perhaps a couple improvements to this proposal could be:
1. Instead of listing "gateway providers" in the `Kuadrant` CR, users could list `GatewayClass` resources;
2. All `Gateway` resources referring `GatewayClasses` listed in the `Kuadrant` CR are to be managed by Kuadrant (cluster admins can always create more `GatewayClasses` to split the sets of gateways that will be managed by Kuadrant from others the won't);
3. Together with each `GatewayClass` referenced in the `Kuadrant` CR, the user would also provide the additional reference to the provider.

> I do not know if it is a real use case, but kuadrant might be configuring multiple gateways managed by different gateway providers at the same time.

I don't think this is possible as definition enforced by the `gatewayClassName`, which always points to one provider.

eguzki:
> I don't think this is possible as definition enforced by the gatewayClassName, which always points to one provider.

I was referring to having multiple gateways with a different class (i.e. different provider). For example, one managed by Istio while other managed by, let's say, Gateway Envoy control plane.

guicassolato:
> > I don't think this is possible as definition enforced by the gatewayClassName, which always points to one provider.
> 
> I was referring to having multiple gateways with a different class (i.e. different provider). For example, one managed by Istio while other managed by, let's say, Gateway Envoy control plane.

Is that possible?

guicassolato:
> > I don't think this is possible as definition enforced by the gatewayClassName, which always points to one provider.
> 
> I was referring to having multiple gateways with a different class (i.e. different provider). For example, one managed by Istio while other managed by, let's say, Gateway Envoy control plane.

My point was that the name of the class itself may not be enough to infer the kind of the provider.

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+
+<br/>
+
+**Alternative B.** Implicit reference to the gateway provider.
+
+With the proposal coming with [RFC Wiring policies with Kuadrant services](https://github.com/Kuadrant/architecture/pull/5), an alternative way to also link the Kuadrant instances and the corresponding control plane/gateway providers into which the Operator needs to inject configuration could be by following the references between either:
+1. the Kuadrant policies (`RateLimitPolicy` and `AuthPolicy`) and their targeted network resources, that ultimately link all the way with gateways and the gateway providers; or
+2. (more directly) the `Kuadrant` custom resources and their corresponding gateway objects, linked via labels and label selectors (see [architecture#5:124-126](https://github.com/Kuadrant/architecture/blob/9545f760b66ff9aaddeeaab126d5276a086e228c/rfcs/0000-wire-policies-with-backend-services.md?plain=1#L124-L126)).
+
+While option 1 above implies a lazy configuration of the functional components into the gateway providers, option 2 perhaps could be implemented in a rather proactive fashion, such as the one proposed in this RFC.
```
maleck13, line 218: not sure I am a fan of the lazy configuration. Or I should say, I like the hands off approach esp for developer exp, but  I think I would want that to be an explicit opt in. It is the role of the  cluster-admin to decide to install kuadrant and it should be up to that persona to decide which gateway providers it applies to. It feels like doing this lazy config moves that reponsibility (perhaps unexpectedly) into the hands of the person defining a RLP for example?

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+N/A.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+1. Is having multiple service mesh control planes (i.e. multiple gateway providers) within the same cluster a real use case or rather just a technical possibility that we choose not to support?<br/>
```
maleck13, line 240: it seems it is a real scenerio  based on recent conversations where there may be two control planes one for the gateway and one for the mesh

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+  **_A:_**<br/><br/>
+2. Does a Kuadrant instance need to be able to target more than one control plane/gateway provider? (See [Rationale and alternatives](#rationale-and-alternative).)<br/>
```
maleck13, line 242: This seems less clear. I see a clear use case for defining which provider an instance of Kuadrant should target, but less so the need for one instance of kuadrant to target multiple providers

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+  **_A:_**<br/><br/>
+3. Should targeting a control plane/gateway provider be required to be explicit (e.g. in a `Kuadrant` CR) or it can based on convention/heuristic (e.g. default to Istio)? _Note:_ I am personally inclined to it needs to be explicit, hence the proposal in this RFC.<br/>
```
maleck13, line 244: I am a fan of explicit configuration for this type of thing over potentially hidden logic

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+  **_A:_**<br/><br/>
+4. What gateway providers should we initially support?<br/>
+  **_A:_** Istio and OpenShift Service Mesh (OSSM).<br/><br/>
+5. What other gateway providers could we aim to support in the future? _Possible answers:_ [Envoy Gateway](https://gateway.envoyproxy.io/v0.2.0/), [Contour](https://projectcontour.io), other API gateways not based on Envoy, ...<br/>
+  **_A:_** Only gateways and perhaps sidecar proxies based on Envoy for quite some time, likely.<br/><br/>
+6. Is "gateway provider" an adequate term? What about supporting injecting Kuadrant into sidecar proxies in the future?<br/>
```
maleck13, line 250: Gateway provider seems adequate a sidcar could be thought of as another form of gateway

rfcs/0000-multiple-control-plane-providers.md
```diff
@@ -0,0 +1,270 @@
...
+  **_A:_** We'll stay with "gateway providers" for now. This can be changed in the future. Sidecars may become a thing one day; this should be either transparent to the scope of this RFC or changeable in the future as part of releasing a new API.<br/><br/>
+7. Should the reconciliation of a `Kuadrant` CR be considered successful only if all gateway providers indicated in the spec were successfuly configured for that instance of Kuadrant? What shall the Operator do in case some of the gateway providers to fail to be configured while others succeed? Shall the Operator roll back the successful configurations in case of partial failure? Should the Operator reflect the state of the reconciliation of each gateway provider listed in the spec in the status subresource of the `Kuadrant` CR? (See [Troubleshooting](#troubleshooting) for possible reasons for the reconciliation to fail.)<br/>
```
maleck13, line 252: Yes I think so and it is probably another good reason to have one provider per instance

maleck13 (commented):
My preference here is to go with explicit configuration and a single provider per kuadrant instance

guicassolato:
Closing this for now this as supporting multiple gateway providers running side by side in the same cluster is no longer an issue. Since [v1.0](https://github.com/Kuadrant/kuadrant-operator/releases/tag/v1.0.0), the Operator will honour the controller behind each GatewayClass for reconciling the internals, whether that controller is Envoy Gateway or Istio.

We can always revisit this proposal in the future in case that selecting a specific provider (to thus exclude others also installed) ever resurfaces as a requirement.

PR #9, opened 2023-02-06
Kuadrant Policies Status conditions/states

didierofrivia:
This RFC presents 2 options and each one could be applied/develop in order and would reflect valuable and accurate information with different degrees of acuity.
Given the feedback so far, it seems Option 2 is the winner:

*Option 2*
Based on [GEP-713](https://gateway-api.sigs.k8s.io/geps/gep-713/#conditions). In this case, besides the proposed Accepted PolicyType, the Enforced PolicyType would be added to reflect the final state of the policy, which means that the policy is showing the synced actual state of the Kuadrant services. The missing Failed PolicyType would be implicitly represented by the TargetNotFound and Invalid PolicyTypeReason.

| Type     | Status | Reason               | Message                                                                     |
|----------|--------|----------------------|-----------------------------------------------------------------------------|
| Accepted | True   | "Accepted"           | "KuadrantPolicy has been accepted"                                          |
|          | False  | "Conflicted"         | "KuadrantPolicy is conflicted by [policy-ns/policy-name], ..."              |
|          | False  | "Invalid"            | "KuadrantPolicy is invalid"                                                 |
|          | False  | "TargetNotFound"     | "KuadrantPolicy target [resource-name] was not found"                       |
| Enforced | True   | "Enforced"           | "KuadrantPolicy has been successfully enforced"                             |
|          | False  | "PartiallyEnforced"  | "KuadrantPolicy has encountered some issues and has been partially applied" |
|          | False  | "Overridden"         | "KuadrantPolicy is overridden by [policy-ns/policy-name], ..."              |


rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,157 @@
...
+* https://github.com/Kuadrant/kuadrant-operator/issues/140
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The following proposal it's divided in 3 stages, where each one could be applied/develop in order and would reflect valuable
+and accurate information with different degrees of acuity. The Policy CRD Status in the following diagrams, are simplified
+as states, which in the [Reference-level explanation](#reference-level-explanation) will be translated to the actual Status Conditions.
+
+## Stage 1
```
maleck13, line 34: I am struggling to visualise this in the resource. Is validiation a status condition type?
IE ```
conditions:
    - lastTransitionTime: "2019-10-22T16:29:24Z"
      status: "True"
      message: "object change observed"
      type: Accepted
    - lastTransitionTime: "2019-10-22T16:29:24Z"
      status: "True"
      message: "change has successfully been applied"
      type: Applied
```

Comment (maleck13) [reply rfcs/0000-rlp-status.md:34]: nvm I see it outlined below

File: rfcs/0000-rlp-status.md [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+
+A simplified version and more aligned with the Kubernetes objects implementation could be represented as the following.
+
+**Conditions**
+
+All conditions are top-level.
+
+| Type        | Status | Reason                      | Message                                                    |
+|-------------|--------|-----------------------------|------------------------------------------------------------|
+| Progressing | True   | "PolicyCreated"             | "RateLimitPolicy created"                                  |
```
maleck13, line 110: I like this simpler model. From this doc I feel it would read better.

Food for thought. I wondered if instead we need two types only

`protected` and `unprotected` ?

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,157 @@
...
+[prior-art]: #prior-art
+
+* [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)
+* [Current RateLimitPolicy Status work](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimitpolicy_status.go)
+
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Should this proposal extend to the KAP as well?
```
maleck13, line 151: I think where possible if we can align the APIs it would make sense from a user exp point of view. Also interesting to think about KAP in the context of `protected` `unprotected`

maleck13 (commented):
Generally like the changes proposed. Although I would tend towards the simplified version and if it makes sense perhaps simplify a little more.

maleck13 (commented):
what are we doing with older RFCs like this, do we leave them open, is it still relevant?

alexsnaps:
> what are we doing with older RFCs like this, do we leave them open, is it still relevant?

We discussed this in last week's planning meeting, we need to do a "update" pass on this one. Make sure all is still in alignment with the v2 policy efforts…

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,162 @@
...
+**The state is defined by the application and validation of the Policy CR**
+
+This first stage is a simple version where the operator only relies on itself, not checking the healthiness of its services
+and just validating the Spec.
+
+![](0000-rlp-status-assets/rlp_status_1.png)
+
+States rationale:
+
+* `Created`: The initial state. It announces that the policy has successfully being created, the operator acknowledges it.
```
maleck13, line 45: have we considered using the same types as Gateway API 

``` Accepted ``` , ```Programmed``` for these states?

Comment (maleck13) [reply rfcs/0000-rlp-status.md:45]: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayStatus

File: rfcs/0000-rlp-status.md [hunk 5]
```diff
@@ -0,0 +1,162 @@
...
+|            | False  | "PolicyNotEnforced"       | "KuadrantPolicy has encountered an error and can't be applied"              |           |
+
+
+A simplified version and more aligned with the Kubernetes objects implementation could be represented as the following.
+
+**Conditions**
+
+All conditions are top-level.
+
+| Type        | Status | Reason                      | Message                                                                     |
```
maleck13, line 111: I think less types would help and it is a good idea to align with K8s or as mentioned above the gateway conditions

rfcs/0000-rlp-status.md
didierofrivia: It's definitely worth considering, I'll update the RFC adding the Gateway Status like alternative

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,162 @@
...
+|             | True   | "PolicyUpdated"             | "KuadrantPolicy has been updated"                                           |
+|             | True   | "PolicyApplied"             | "KuadrantPolicy has been successfully applied                               |
+|             | True   | "PolicyReconciled"          | "KuadrantPolicy has been successfully reconciled"                           |
+|             | False  | "PolicyEnforced"            | "KuadrantPolicy has been successfully enforced"                             |
+|             | False  | "PolicyError"               | "KuadrantPolicy has encountered an error"                                   |
+| Enforced    | True   | "PolicyEnforced"            | "KuadrantPolicy has been successfully enforced"                             |
+|             | False  | "PolicyPartiallyEnforced"   | "KuadrantPolicy has encountered some issues and has been partially applied" |
+|             | False  | "PolicyNotEnforced"         | "KuadrantPolicy has encountered an error and can't be applied"              |
```
guicassolato, line 121: Is there an overlap between `Enforced: False (PolicyNotEnforced)` and `Failed: True (PolicyReconciliationError)` maybe?

What about specifying a more specific reason why the policy is not enforced? E.g.

```suggestion
|             | False  | "PolicyOverridden"         | "KuadrantPolicy overridden by KuadrantPolicy policy-ns/policy-name"              |
```

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,162 @@
...
+| Failed      | True   | "PolicyValidationError"     | "KuadrantPolicy has failed to validate"                                     |
+|             | True   | "PolicyReconciliationError" | "KuadrantPolicy has encountered a reconciliation error"                     |
+|             | True   | "PolicyServiceError"        | "KuadrantPolicy has encountered has failed to enforce"                      |
+|             | False  | "PolicyEnforced"            | "KuadrantPolicy has been successfully enforced"                             |
```
guicassolato, line 125: Is the policy being overridden by another policy to be considered a fail? If it's not a fail, then we may need another reason here.

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The following proposal it's divided in 3 stages, where each one could be applied/develop in order and would reflect valuable
+and accurate information with different degrees of acuity. The Policy CRD Status in the following diagrams, are simplified
+as states, which in the [Reference-level explanation](#reference-level-explanation) will be translated to the actual Status Conditions.
+
+## Stage 1
+**The state is defined by the application and validation of the Policy CR**
```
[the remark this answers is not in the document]
didierofrivia: Yes, it basically is about the number 4 you mention above. I didn't mind anything below the admission control stage, but if we're gonna have our own set of rules for admission control, it might reflect this in the state

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,162 @@
...
+
+This first stage is a simple version where the operator only relies on itself, not checking the healthiness of its services
+and just validating the Spec.
+
+![](0000-rlp-status-assets/rlp_status_1.png)
```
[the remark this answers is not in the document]
didierofrivia: That is correct and more clear now with Policy V2, regarding the indirect changes

maleck13:
@didierofrivia this section of the policy attachment gep may be relevant https://gateway-api.sigs.k8s.io/geps/gep-713/#conditions

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,162 @@
...
+|             | False  | "PolicyError"               | "KuadrantPolicy has encountered an error"                                   |
+| Enforced    | True   | "PolicyEnforced"            | "KuadrantPolicy has been successfully enforced"                             |
+|             | False  | "PolicyPartiallyEnforced"   | "KuadrantPolicy has encountered some issues and has been partially applied" |
+|             | False  | "PolicyNotEnforced"         | "KuadrantPolicy has encountered an error and can't be applied"              |
+| Failed      | True   | "PolicyValidationError"     | "KuadrantPolicy has failed to validate"                                     |
+|             | True   | "PolicyReconciliationError" | "KuadrantPolicy has encountered a reconciliation error"                     |
+|             | True   | "PolicyServiceError"        | "KuadrantPolicy has encountered has failed to enforce"                      |
+|             | False  | "PolicyEnforced"            | "KuadrantPolicy has been successfully enforced"                             |
+
+
```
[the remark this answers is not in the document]
didierofrivia: This is definitely a good thing for a dashboard or an aggregated data resource that provides more insight about the Kuadrant policies overall state. Implementation wise, I can't foresee how possible it is to implement it per policy state, however, we could include some kind of brief informative summary or list of conflicting policies, network resources affected, etc...

guicassolato (commented):
Although I'm not sure if we can model those exact state transitions depicted in the diagrams of the 3 stages, or if they serve rather as a guide to explain the general lifecycle of a policy (i.e., at documentation level only), I am quite happy with **Option 3** proposed, and would love to move forward with that.

Another thing to watch in relation to this is how Gateway API's proposal for a [`PolicyBinding` resource](https://gateway-api.sigs.k8s.io/geps/gep-713/#policybinding-resource) will evolve (and how _we_ can contribute to that process!) Reflecting everything that may have happened only in the policy status sub-resource may end up simply not being feasible. Hence, if an alternative such as a policy binding resource (or kubectl plugin, or CLI tool, etc) exists, the policy conditions and reasons can be kept simple and high level, much along the lines of what's proposed here.

Nice job, @didierofrivia!

rfcs/0000-rlp-status.md
```diff
@@ -0,0 +1,203 @@
...
+All conditions are top-level.
+
+| Type           | Status | Reason              | Message                                                                     |
+|----------------|--------|---------------------|-----------------------------------------------------------------------------|
+| Accepted       | True   | "Accepted"          | "KuadrantPolicy has been accepted"                                          |
+|                | False  | "Conflicted"        | "KuadrantPolicy is conflicted by [policy-ns/policy-name], ..."              |
+|                | False  | "Invalid"           | "KuadrantPolicy is invalid"                                                 |
+|                | False  | "TargetNotFound"    | "KuadrantPolicy target [resource-name] was not found"                       |
+| Enforced       | True   | "Enforced"          | "KuadrantPolicy has been successfully enforced"                             |
+|                | False  | "PartiallyEnforced" | "KuadrantPolicy has encountered some issues and has been partially applied" |
```
maleck13, line 155: Not sure I like partially enforced. might be tempted to group that under reason `Unknown` this is similar then to how the gateways themselves reflect transient or uncertain state.

maleck13 (commented):
I think we should focus around option 2 that reflects the GEP status concepts. @didierofrivia does it makes sense to update this with a "ruled out" section to keep a record of the other options but simplify the overall doc?

alexsnaps (approved):

rfcs/0000-rlp-status.md
didierofrivia: Since we can see the information of partially enforced within the `PolicyAffected` type, let's change this one to `Unknown`

PR #16, opened 2023-05-02
Custom underlying services

eguzki:
Related issue: https://github.com/Kuadrant/kuadrant-operator/issues/164

### DIFF
--- a/rfcs/0000-custom-underlying-services.md
+++ b/rfcs/0000-custom-underlying-services.md
@@ -42,7 +42,7 @@ spec:
 The **target references** are themselves **custom resources** representing an instance of the service. Specifically:
 
 * `authorinoRef` contains a reference to [Authorino CR](https://github.com/Kuadrant/authorino-operator/blob/6e7b0005a35107aafc01f2cfb7d7e235ea237bbf/config/crd/bases/operator.authorino.kuadrant.io_authorinos.yaml) provided by the [Authorino Operator](https://github.com/Kuadrant/authorino-operator).
-* `limitadorRef` contains a reference to one [Limitador CR](https://github.com/Kuadrant/limitador-operator/blob/d68a13c12a5ab76cc93547a8433a49cfef92f18c/api/v1alpha1/limitador_types.go) provided by the [Limtiador Operator](https://github.com/Kuadrant/limitador-operator)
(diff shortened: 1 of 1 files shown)


eguzki:
There are still some `TODO`s and empty sections. Looking for some first comments, though.

rfcs/0000-custom-underlying-services.md
```diff
@@ -0,0 +1,163 @@
...
+- It is reasonably clear how the feature would be implemented.
+- How error would be reported to the users.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+TODO
```
pehala, line 106: One potential drawback is that we do not want users to change some settings. From the top of my head, I can think of Authorino `spec.clusterWide` and `authConfigLabelSelectors`, which could mean that the user could easily end up with AuthPolicy that will be reconciled by Kuadrant but won't be picked up in Authorino.

rfcs/0000-custom-underlying-services.md
```diff
@@ -0,0 +1,163 @@
...
+One kuadrant control plane instance is represented by one Kuadrant custom resource. The main idea is to wire the underlaying services backing kuadrant with the kuadrant instance within the Kuadrant custom resource spec.
+
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+  name: <kuadrant-name>
+  namespace: <kuadrant-namespace>
+spec:
+  authorinoTargetRef:
```
guicassolato, line 32:
```suggestion
  authorinoRef:
```

rfcs/0000-custom-underlying-services.md
```diff
@@ -0,0 +1,163 @@
...
+      name: <authorino-name>
+      namespace: <authorino-namespace>
+  limitadorTargetRef:
```
guicassolato, line 35:
```suggestion
  limitadorRef:
```

rfcs/0000-custom-underlying-services.md
eguzki: Kuadrant should reconcile those changes in the external services, right? More specifically, when the Authorino CR is (externally) updated and the label selectors are updated, the AuthConfig CR's managed by kuadrant should reconcile to match.

eguzki:
Updating with my latest changes and handing over to anyone from the team

PR #18, opened 2023-07-08
Shipping with an embedded GW API provider

alexsnaps:
Trying to lower the bar to entry into Kuadrant… if only until all k8s clusters out there have Gw API installed (using Envoy) 😜

### DIFF
--- a/rfcs/0000-envoy-gw.md
+++ b/rfcs/0000-envoy-gw.md
@@ -1,0 +1,98 @@
+# RFC Template
+
+- Feature Name: envoy-gw
+- Start Date: 2023-07-07
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Provide (different?) deployment for Kuadrant, either with a default Gateway API provider or a pluggeable
+backend. The proposal is to provide users a battery included experience, when getting started with
+Kuadrant while not yet having a (supported) Gateway API provider installed on their cluster.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, Kuadrant comes with no batteries included, i.e. no Gateway API provider. This leaves it up to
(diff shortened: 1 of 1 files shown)


rfcs/0000-envoy-gw.md
```diff
@@ -0,0 +1,98 @@
...
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Doing so entails a few things: 
+
+ - [ ] Get an _Envoy Gateway_ distribution ready to back Kuadrant use cases
+   - including _howtos_ on how to "chain" existing ingress'es with gateway resources
+ - [ ] Abstract the Gateway Provider behind an [Service Provider interface (SPI)](https://en.wikipedia.org/wiki/Service_provider_interface), to support at least Envoy GW, Istio & OSSM.
+   - which makes for a clear contract any other Gateway API provider would be able to implement
+ - [ ] Package different distributions (at the very least GW API included, or pluggeable) as different Operators/images?
```
maleck13, line 47: so would the idea here be to support go plugins https://pkg.go.dev/plugin and have the provider configured at install time a loaded at dynamically? Would our default impl also be plugins or just packaged as part of the operator?

alexsnaps: I'd say that'd depend on "how big" these become… And how we'd even plug the default Gateway API provider (i.e. probably Envoy Gateway) in. I think, from a end-user's perspective, I'd lean towards one single operator image that just "does the right" thing. No Gateway API present? Go fetch and install our default provider; Or, we;re in an OSSM environment? Loads the appropriate SPI (possibly using the go plugin - which I guess are effectively a `dlopen(3)` kinda thing; or "just" using a factory indirection; or whichever…) and it all works.
I'll look into go plugins, but at this stage I mostly care about the definition of the contract between Kuadrant and these providers, to be defined in this/these SPI(s). And if we can then also easily accept contribution to support "yet another Gateway API provider", then great! If that is more complicated, supporting go plugins would be another way to decouple Kuadrant from the providers, all the way to the runtime, creating an even more open architecture for these extensions.

guicassolato:
👍 to the intro written so far.

maleck13:
Is this still relevant @alexsnaps . We have project sail as an option but wondering do you see more to do here?

PR #23, opened 2023-08-25
fix: rfc 0002 filename and internal refs

guicassolato:


didierofrivia (approved):
Was about to mention this hehe 👍🏼

PR #25, opened 2023-09-11
rfc: configuration of kuadrant sub components

Boomatang:
Address needs of https://github.com/Kuadrant/kuadrant-operator/issues/163
Related issue: https://github.com/Kuadrant/kuadrant-operator/issues/164

### DIFF
--- a/rfcs/0000-kuadrant_sub_components_configurations.md
+++ b/rfcs/0000-kuadrant_sub_components_configurations.md
@@ -1,0 +1,237 @@
+# Configuration of Kuadrant Sub Components
+
+- Feature Name: Configuration of Kuadrant Sub components.
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/kuadrant-operator#163](https://github.com/Kuadrant/kuadrant-operator/issues/163)
+
+# Summary
+[summary]: #summary
+
+Enable configuration of sub components of Kuadrant from a centralized location, namely the Kuadrant CR. 
+
+# Motivation
+[motivation]: #motivation
+
+The initial request comes from MGC to configure Redis for Limitador by the following issue [#163](https://github.com/Kuadrant/kuadrant-operator/issues/163).
+MGCs current work around is to update the Limitador CR after the deployment with the configuration setting for Redis Instance.
+This change would allow for the configuration sub components before the Kuadrant is deployed. 
+
+As a kuadrant user this reduces the number of CRs that they are required to modify to get the installation they require.
+The sub components CRDs (Authorino, Limitador) never have to be modified by a Kuadrant user (and should never be modified by a Kuadrant User).
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+As the Kuadrant operator would be responsible for reconciling these configurations into the requested components, restrictions and limitations can be placed on the components which maybe allowed in a standalone installation.
+An example in this space is the disk storage for Limitador which is a new feature and the Kuadrant installation may not want to support it till there is a proven track record for the feature.
+
+For existing Kuadrant Users this may be a possible breaking changes if those users manually configure the Kuadrant sub components via their CRs.
+A guide can be created to help migrate the users configurations to the Kuadrant CR.
+This guide can be part of the release notes and/or possibly released before the release of Kuadrant.
+
+The deployment configuration for each component can be placed in the Kuadrant CR.
+These configurations are then reconciled into the CRs for each component. 
+Only the options below are exposed in the Kuadrant CR. 
+All fields in the spec are optional.
+
+``` yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+    name: kuadrant-sample
+spec:
+    limitador:
+        afffinity: ...
+        listener: ...
+        pdb: ...
+        replicas: ...
+        resourceRequirements: ...
+        storage: ...
+    authorino:
+        eveluatorCacheSize: ...
+        healthz: ...
+        listener: ...
+        logLevel: ...
+        metrics: ...
+        oidcServer: ...
+        replicas: ...
+        tracing: ...
+        volumes: ...
+status:
+    ...
+```
+
+The Kuadrant operator will watch for changes in the Authorino and Limitador CRs, reconciling back any changes that a user may do to these configurations.
+How ever Kuadrant operator will not reconcile fields that are given above. 
+An example of this is the `image` field on the Authorino CR.
+This field allows a user to set the image that Authorino is deployed with.
+The feature is meant for dev and testing purposes. 
+If a user wishes to use a different image, they can.
+Kuadrant assumes they know what they are doing but requires the user to set the change on the component directly.
+
+Only the sub component operator will be responsible for actioning the configurations pasted form the Kuadrant CR to the sub components CR.
+This ensure no extra changes will be required in the sub operators to meet the needs of Kuadrant.
+
+Status errors related to the configuration of the sub components should be reported back to the Kuadrant CR.
+The errors messages in Kuadrant state what components are currently having issue and which resource to review for more details.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+All the fields in the Authorino and Limitador CRs that are configurable in the Kuadrant CR are optional and have sound defaults.
+Kuadrant needs to remain installable with out having to set any spec in the Kuadrant CR.
+
+The Kuadrant operator should only reconcile the spec that is given. 
+This would mean if the user states the number of replicas to be used in one of the components only the replica field for that component should be reconciled.
+As the other fields would be blank at this stage, blank fields would not be reconciled to the component CR.
+By this behaviour a few things are being achieved.
+Component controllers define the defaults to be used in the components.
+Optional fields in the component CRs never get set with blank values.
+Blank values in the component CR could override the defaults of the components causing unexpected behaviour.
+Existing Kuadrant users may already have custom fields set in the component CRs.
+By only reconciling the fields set in the kuadrant CR this allows time for a user to migrate their custom configuration from the component CR to the Kuadrant CR.
+
+## Fields to reconcile
+Fields being reconcile can be classified into different groups.
+These classifications are based around the tasks a user is achieve.
+
+- Kubernetes native, setting that affect how Kubernetes handles the resource.
+- Observability, configuration settings that allow insights into how the applications are operation.
+This can be Kubernetes native or external tooling.
+- Application Settings, setting targeting the application and how it connects to external services.
+
+### Authorino Spec
+#### Kubernetes native
+- replicas
+
+#### Observability
+- healthz
+- logLevel
+- metrics
+- tracing
+
+#### Application Settings
+- eveluatorCacheSize
+- listener
+- oidcServer
+- volumes
+
+### Limitador Spec
+#### Kubernetes native
+- afffinity
+- pdb
+- replicas
+- resourceRequirements
+
+#### Application Settings
+- listener
+- storage
+
+## Fields not reconciled
+There are a number of fields in both Authorino and Limitador that are not reconciled.
+Reasons for doing this are:
+
+It is better to start with a sub set of features and expand to include more at a later date.
+Removing feature support is far harder than adding it.
+
+There are four classifications the unreconciled fields fail into.
+- Deprecated, fields that are deprecated and/or have plans to be removed from the spec in the future.
+- Unsupported, the features would have hard coded or expected defaults in the Kuadrant operator. 
+Work would be required to all the custom configurations options.
+- Dev/Testing focused, features that should only be used during development & testing and not recommended for production.
+The defaults would for the fields are the production recommendations.
+- Reconciled by others, this mostly affects Limitador as the deployment configuration and runtime configuration are in the same CR.
+In the case of Kuadrant the runtime configuration for Limitador is added via the RateLimitingPolicy CR.
+
+### Authorino Spec
+
+#### Unsupported
+- clusterWide
+- authConfigLabelSelectors
+- secrtLabelSelectors
+
+#### Dev/Testing focused
+- image
+- imagePullPolicy
+- logMode
+
+### Limitador Spec
+
+#### Unsupported
+- RateLimitHeaders
+
+#### Reconciled by others
+- Limits
+
+#### Deprecated
+- version
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+As the Kuadrant CR spec will be a sub set of the features that can be configured in the sub components spec, extra maintenances will be required to ensure specs are in sync.
+
+New features of a component will not be accessible in Kuadrant initially.
+This is both a pro and a con.
+
+Documentation becomes harder, as the sub component should be documenting their own features but in Kuadrant the user does not configure the feature in sub component.
+This has the risk of confusing new users.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+
+One alternative that was being looked at was allowing the user to bring their own Limitador instances by stating which Limitador CR Kuadrant should use. 
+A major point of issue with this approach was knowing what limits the user had configured and what limits Kuadrant configured. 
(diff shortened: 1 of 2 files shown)


alexsnaps:
This might be of interest to this:
https://github.com/Kuadrant/limitador-operator/pull/96

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,248 @@
...
+In the case of Kuadrant the runtime configuration for Limitador is added via the RateLimitingPolicy CR.
+
+### Authorino Spec
+
+#### Unsupported
+- clusterWide
+- authConfigLabelSelectors
+- secrtLabelSelectors
+
+#### Dev/Testing focused
```
pehala, line 154: I would actually include those as well but with a clear indication that it should not be used for production. It might make debugging or trying out newer Authorino versions simpler.

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,248 @@
...
+These classifications are based around the tasks a user is achieve.
+
+- Kubernetes native, setting that affect how Kubernetes handles the resource.
+- Observability, configuration settings that allow insights into how the applications are operation.
+This can be Kubernetes native or external tooling.
+- Application Settings, setting targeting the application and how it connects to external services.
+
+### Authorino Spec
+#### Kubernetes native
+- replicas
```
pehala, line 106: Is there a reason why the Authorino k8s native does not include `affinity`, `pdb` and `resourceRequirements`?

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,248 @@
...
+- Does another project have a similar feature?
+- What can be learned from it? What's good? What's less optimal?
+- Papers: Are there any published papers or great posts that discuss this? If you have some relevant papers to refer to, this can serve as a more detailed theoretical background.
+
+This section is intended to encourage you as an author to think about the lessons from other tentatives - successful or not, provide readers of your RFC with a fuller picture.
+
+Note that while precedent set by other projects is some motivation, it does not on its own motivate an RFC.
+
+------
+In RHOAM a user can configure the custom SMTP and custom domains for 3scale from the addon forms.
```
pehala, line 210: Would it be possible to not use RHOAM as an example? I think it is too niche of an example to make sense for most users (e.g. Red Hat external contributors)

rfcs/0000-kuadrant_sub_components_configurations.md
Boomatang: Currently Authorino-operator does not yet support configuring the `affinity`, `pdb` or `resourceRequirements` resources. To my knowledge no one has requested the feature yet. The limitador-operator only got this feature in the last few months. 

If and when Authorino-operator is given this feature I would expect the fields to bubble up to the Kuadrant CR. Should Authorino-operator get these feature before this RFC is implemented, that is a different question. Personal I don't think it needs to but would be nice to have.

rfcs/0000-kuadrant_sub_components_configurations.md
Boomatang: I am in two minds about this, yes it would make using newer version of Authorino easier but also people don't read the documentation so how do you give a clear indication it should not be used in production. By making the route to make such changes harder it is an indication that a user should not being doing it. At no point am I suggesting that we don't allow the setting of these fields but just that it should be done it the sub operator directly. 

Take the example  of the `image` field. it allows the setting of a custom Authorino image. If this is set in the Kuadrant CR the custom image will be set during the installation of Kuadrant. When the image is set in the Authorino CR the Authpolicies have already being created. This would be a very deliberate act need to use a custom image.  

If the customer is using GitOps to manage the resources setting a custom image in the Kuadrant CR is very simple but also setting it the image in Authorino would be possible, it needs a more advance set up which should be an indication that this might not be the correct way. 

When it comes to a customer using custom images and pull policies I would suspect they are also the kind of customer that does not use auto update and well use to having extra steps to ensure the their configuration.

rfcs/0000-kuadrant_sub_components_configurations.md
Boomatang: I don't mind removing this, but I don't have a good alternative example. I did not think of the example from a external contributors point of view.

rfcs/0000-kuadrant_sub_components_configurations.md
pehala: And will we allow setting things in suboperator, or is it expected that Kuadrant reconciles back all of the changes?

Boomatang: Yes, if the field can not be set in the Kuadrant CR the kuadrant-operator should not reconcile back any changes to that field. This allow the configuration of the test/dev features directly on the suboperator. A side affect of this would mean any new feature in the suboperators do not need to be in the Kuadrant CR for the user to make use of the feature. This would be like a soft launch.

Thinking about what this means for upgrades. If the kuadrant-operator reverts the configuration on the suboperators to the configuration in the Kuadrant CR this could cause issues for the end users. But also can be documented as an upgrade step/requirement. I believe this would be an acceptable step.

pehala: Ok, that makes sense to me

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,248 @@
...
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to resolve through the RFC process before this gets merged?
+- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+
+-----
+- Is there a need to add validation on the configuration?
+- If a valid configuration is add to the Kuadrant CR and this configuration is pass to the sub components CR but there is a error trying to setting up the configuration.
+How is this error reported back to the user?
```
pehala, line 230: I would say the Kuadrant should watch the component CRs and then pass the error from the component CR to the Kuadrant CR. Kuadrant CR should be ready only when all components CRs are ready.

Boomatang: Agreed the Kurdant CR should only report ready when all component CRs are ready. Should the component error be past up to the Kuadrant CR or should the Kuadrant CR report that the component CR is in error or not ready state?

Passing the component error up could be messy if the components don't use the same format for the errors. Stating a component is in an error state can be uniformed but requires the use the user to do an extra hop.

pehala: I would rather have the error present in the Kuadrant CR. I think the tradeoff is worth it.

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+- Kubernetes native, setting that affect how Kubernetes handles the resource.
+- Observability, configuration settings that allow insights into how the applications are operation.
+This can be Kubernetes native or external tooling.
+- Application Settings, setting targeting the application and how it connects to external services.
+
+### Authorino Spec
+#### Kubernetes native
+- replicas
+
+#### Observability
```
maleck13, line 108: what am I setting when configuring these? LogLevel is obvious, but what is the value to exposing Healthz, Metrics, Tracing ? Should these be things we set as part of the kaudrant install. I guess I am struggling to see the use case for setting some of these and perhaps we should limit down the set we want to expose further @alexsnaps ?

maleck13 (commented):
I think it is reasonable to allow some of the configuration options to bubble up to the kuadrant CR. I have concerns with how many we are proposing. I am not sure we have clear use case for each of them. The redis url has a clear use case for example, but I don't know why I am setting the number replicas for example (why can't I just scale up the deployment) this seems more complex than what is needed. I worry a little about leaking things that exist in established APIs such as deployments up through not one but two CRs. Personally I have become more of a fan of you get the default deployment and there are things you can chane about that if you want (IE numbr of replicas)

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
+# Configuration of Kuadrant Sub Components
+
+- Feature Name: Configuration of Kuadrant Sub components.
```
guicassolato, line 3:
```suggestion
- Feature Name: `sub-components-config`
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+# Summary
+[summary]: #summary
+
+Enable configuration of sub components of Kuadrant from a centralized location, namely the Kuadrant CR. 
+
+# Motivation
+[motivation]: #motivation
+
+The initial request comes from MGC to configure Redis for Limitador by the following issue [#163](https://github.com/Kuadrant/kuadrant-operator/issues/163).
+MGCs current work around is to update the Limitador CR after the deployment with the configuration setting for Redis Instance.
```
guicassolato, line 17:
```suggestion
MGC's current work around is to update the Limitador CR after the deployment with the configuration setting for Redis Instance.
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+This change would allow for the configuration sub components before the Kuadrant is deployed. 
+
+As a kuadrant user this reduces the number of CRs that they are required to modify to get the installation they require.
+The sub components CRDs (Authorino, Limitador) never have to be modified by a Kuadrant user (and should never be modified by a Kuadrant User).
```
guicassolato, line 21:
```suggestion
The sub components CRs (Authorino, Limitador) never have to be modified by a Kuadrant user (and should never be modified by a Kuadrant User).
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+[summary]: #summary
+
+Enable configuration of sub components of Kuadrant from a centralized location, namely the Kuadrant CR. 
+
+# Motivation
+[motivation]: #motivation
+
+The initial request comes from MGC to configure Redis for Limitador by the following issue [#163](https://github.com/Kuadrant/kuadrant-operator/issues/163).
+MGCs current work around is to update the Limitador CR after the deployment with the configuration setting for Redis Instance.
+This change would allow for the configuration sub components before the Kuadrant is deployed.
```
guicassolato, line 18:
```suggestion
This change would allow for the configuration of sub components before the Kuadrant is deployed. 
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+Enable configuration of sub components of Kuadrant from a centralized location, namely the Kuadrant CR. 
+
+# Motivation
+[motivation]: #motivation
+
+The initial request comes from MGC to configure Redis for Limitador by the following issue [#163](https://github.com/Kuadrant/kuadrant-operator/issues/163).
+MGCs current work around is to update the Limitador CR after the deployment with the configuration setting for Redis Instance.
+This change would allow for the configuration sub components before the Kuadrant is deployed. 
+
+As a kuadrant user this reduces the number of CRs that they are required to modify to get the installation they require.
```
guicassolato, line 20:
```suggestion
This reduces the number of CRs that users of Kuadrant are required to modify to get the installation they require.
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+- replicas
+
+#### Observability
+- healthz
+- logLevel
+- metrics
+- tracing
+
+#### Application Settings
+- eveluatorCacheSize
```
guicassolato, line 115:
```suggestion
- evaluatorCacheSize
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+The defaults would for the fields are the production recommendations.
+- Reconciled by others, this mostly affects Limitador as the deployment configuration and runtime configuration are in the same CR.
+In the case of Kuadrant the runtime configuration for Limitador is added via the RateLimitingPolicy CR.
+
+### Authorino Spec
+
+#### Unsupported
+- clusterWide
+- authConfigLabelSelectors
+- secrtLabelSelectors
```
guicassolato, line 152:
```suggestion
- secretLabelSelectors
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+
+The Kuadrant operator will watch for changes in the Authorino and Limitador CRs, reconciling back any changes that a user may do to these configurations.
+How ever Kuadrant operator will not reconcile fields that are given above. 
+An example of this is the `image` field on the Authorino CR.
+This field allows a user to set the image that Authorino is deployed with.
+The feature is meant for dev and testing purposes. 
+If a user wishes to use a different image, they can.
+Kuadrant assumes they know what they are doing but requires the user to set the change on the component directly.
+
+Only the sub component operator will be responsible for actioning the configurations pasted form the Kuadrant CR to the sub components CR.
```
guicassolato, line 73:
```suggestion
Only the sub component operator will be responsible for actioning the configurations pasted from the Kuadrant CR to the sub components CR.
```

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+Optional fields in the component CRs never get set with blank values.
+Blank values in the component CR could override the defaults of the components causing unexpected behaviour.
+Existing Kuadrant users may already have custom fields set in the component CRs.
+By only reconciling the fields set in the kuadrant CR this allows time for a user to migrate their custom configuration from the component CR to the Kuadrant CR.
+
+## Fields to reconcile
+Fields being reconcile can be classified into different groups.
+These classifications are based around the tasks a user is achieve.
+
+- Kubernetes native, setting that affect how Kubernetes handles the resource.
```
guicassolato, line 99: I wonder about labels. Are those in this group? Or maybe they could be propagated from the labels of the Kuadrant CR itself. If this is the case, are labels out of scope of the RFC?

The ability to set labels in the components' CRs can be useful for selecting not only these custom resources in particular, but also other resources that the corresponding component operator creates based on those. This is the case of Authorino, whose operator creates services propagating the same labels applied to the Authorino CR, including, e.g., Authorino's metrics service, later used by the scraper to find the service.

rfcs/0000-kuadrant_sub_components_configurations.md
Boomatang: What is the reasoning behind this change?

rfcs/0000-kuadrant_sub_components_configurations.md
Boomatang: Labels are interesting, and would be out of scope of the RFC. This RFC is focusing on  the spec of CRs. 

If the labels were to be managed you would also need to take into account other updater's of the labels on the resources. The source of truth for the labels becomes a bit muddy.

maleck13:
I am happy to approve as long as we give careful consideration to each value we choose to surface.

rfcs/0000-kuadrant_sub_components_configurations.md
```diff
@@ -0,0 +1,237 @@
...
+spec:
+    limitador:
+        afffinity: ...
+        listener: ...
+        pdb: ...
+        replicas: ...
+        resourceRequirements: ...
+        storage: ...
+    authorino:
+        eveluatorCacheSize: ...
```
pehala, line 52:
```suggestion
        evaluatorCacheSize: ...
```

rfcs/0000-kuadrant_sub_components_configurations.md
guicassolato: Following the pattern from other RFCs. I think the idea for this is to be like a unique identifier that can be used to track the feature unequivocally across, e.g., file names, etc. It helps avoid overloading words and related confusion of such of when relying otherwise only on the human version of it (which is already in the title, BTW.)

rfcs/0000-kuadrant_sub_components_configurations.md
guicassolato: One thing is for sure. We need a way for users to be able to propagate labels across resources initially created by our controllers.

That said, I can live with labels being out of the scope of this RFC. It should be clear then for one who implements this RFC that reconciling the resources back to the desired state declared in the `Kuadrant` CR should NOT include the labels (at least not because of this RFC.)

guicassolato (approved):

PR #37, opened 2023-11-02
Multi-threaded Limitador

didierofrivia:


rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,237 @@
...
+### Configuration and mode selection
+
+The configuration of the service will be done at the initialization time, and it will be done through the command line as
+described in the [Guide-level explanation](#guide-level-explanation) section.
+
+#### Accuracy mode
+
+* In Accuracy mode, we need to strictly adhere to the defined limits by denying requests that exeed the limits for each
+Counter Object.
+* Check and update Counter values atomically to maintin consistency and prevent race conditions.
```
adam-cattermole, line 185:
```suggestion
* Check and update Counter values atomically to maintain consistency and prevent race conditions.
```

rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,237 @@
...
+* Enforce rate limits accurately by comparing the number of hits within the sliding window to the defined limit.
+
+#### Throughput mode
+
+* In Throughput mode, we need to prioritize maximizing the number of requests that can be processed concurrently over
+strict adherence to defined limits.
+* Allowing request to proceed even if they temproarly exceed the defined limits, favouring higher request rates.
+* We should use appropriate concurrency mechanisms and CAS oeprations to handle request efficiently while maintaining
+consistency in the counter data. We might not need to use locks, but we need to be careful with the CAS operations nonetheless.
```
adam-cattermole, line 194:
```suggestion
* Allowing request to proceed even if they temporarily exceed the defined limits, favouring higher request rates.
* We should use appropriate concurrency mechanisms and CAS operations to handle requests efficiently while maintaining
consistency in the counter data. We might not need to use locks, but we need to be careful with the CAS operations nonetheless.
```

rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,241 @@
...
+
+# Summary
+[summary]: #summary
+
+Enable Limitador service to process requests in parallel when configured to use In Memory storage.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, Limitador service is single threaded, regardless of its chosen storage. This means that it can only process one request at a time.
```
eguzki, line 15: This is actually not true, is it? The gRPC service (tonic) is built on top of Tokio, an event-driven, non-blocking I/O platform. Or when the storage is memory, there is no I/O involved, hence only one request at a time can be processed?

rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,241 @@
...
+introduce an interface that allows the user to meet their requirements.
+
+A couple of concepts we need to define in order to understand the proposal:
+* **Accuracy**: The _accuracy_ is defined by the service's ability to enforce the limits in an exact factual way. This means
+  that if the service is configured to allow 10 requests per second, it will only allow 10 requests per second, and
+  not 11 or 9. In terms of observability, the counters, will be monotonically strictly decreasing without repeating any
+  value given per service.
+* **Throughput**: The _throughput_ of a service is defined by the number of requests that the service can process in a
+  given time. As a consequence of having higher throughput, we need to introduce two more concepts:
+  * **Overshoot**: The _overshooting_ of a limit counter is defined by the difference between the _expected_ value of the counter
```
eguzki, line 34: _overshooting_ is when the counter goes beyond the limit and the request is still admitted, right?

rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,241 @@
...
+    and the value that the counter has in the storage, when the stored value is **greater** than the expected one.
+  * **Undershoot**: The _undershooting_ of a limit counter is defined the same way the **Overshoot** is, but the stored value
+    is **lower** than the expected one.
+
+## Behaviour
+
+When using the multi-threading approach, we need to understand the trade-offs that we are making. The main one is that
+it's not possible to have both _Accuracy_ and _Throughput_ at the same time. This means that we need to choose one of
```
eguzki, line 42: it would be nice to explain why both are not achievable.

rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,241 @@
...
+conditions:
+  - "req.method == 'POST'"
+```
+While this one would be translated to _POST_ requests to the namespace `example.org` can be authorized up to 2 times
+in a span of 1 second.
+
+Now imagine that we have the following requests happening at the same time in parallel: `POST`, `POST`, `POST`, `GET`,
+`GET`; it could happen that the service authorizes the first two `POST` requests and updates both counters to 2, then the third
+`POST` request is not authorized *but* the counter of limit 1 is updated to 3 (wrongly), and finally then only one `GET` request
+that comes in would be authorized, leaving one wrongly denied. In this case, we would have an _overshoot_ of 1 in the
```
eguzki, line 111: I would say this is a _undershooting_ example, i.e. traffic is being rejected *before* reaching the limit.

eguzki: I may need to review my understanding about these concepts.

rfcs/0000-limitador-multithread-inmemory.md
```diff
@@ -0,0 +1,241 @@
...
+### Request handling and concurrency control
+
+When a request comes in, Limitador needs to determine the appropriate Counter object(s) based on the request's namespace
+and limit definitions. Taking into account the collection will be in ascending order by expiration time, iterating over the
+counter objects associated with those premises and performing the following steps:
+
+1. Retrieve the Counter object from the collection.
+2. Compare the current timestamp with the expiration time of the counter to determine if the request falls within the
+sliding window
+3. If the request falls within the window, increment the hit count of the Counter.
```
eguzki, line 162: shouldn't the increment happen only when it has been verified that no counters has exceeded the associated limit?

eguzki (commented):
I might be wrong, but it seems to me that no one requires 100% accuracy. So `--mode=accuracy` would not be necessary. For me it is more about what level of the rate `throughput / accuracy` you want. Having valid values meaningful names, never numbers. Something like 
* `high` => I do not care about accuracy, I want it as fast as possible
* `medium`  =>  I care about both
* `low` => I care about accuracy but assuming it will not be 100% accurate

rfcs/0000-limitador-multithread-inmemory.md
alexsnaps: See [this issues](https://github.com/Kuadrant/limitador/issues/14), other than something having changed in newer version, this was still true like 10 months ago.

rfcs/0000-limitador-multithread-inmemory.md
alexsnaps: Interesting, I agree with your definition, but looks like @didierofrivia looks at it the other way around.

maleck13:
This has been open for quite some time. Is there more to happen with it?

didierofrivia:
@maleck13 will consult with @alexsnaps @adam-cattermole and @eguzki regarding the future plans of limitador and if this is something we still are considering

PR #41, opened 2023-11-21
Kuadrant components CI/CD

didierofrivia:
Partially closes https://github.com/Kuadrant/architecture/issues/59
Aspects regarding the actual event of releasing: https://github.com/Kuadrant/architecture/pull/46


rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,326 @@
...
+quay.io/kuadrant/authorino:<git-ref> – the quay.io/kuadrant/authorino:latest tag is also moved to match the latest <git-ref>.
+
+### Authorino Operator
+The Authorino operator is similar to its operand, it fully follows the [Operator Pattern](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/)
+and to build its manifests relies on [Kustomize](https://kustomize.io/).
+
+#### Artifacts
+Its artifacts are the [Authorino Operator image](https://quay.io/repository/kuadrant/authorino-operator),
+[Authorino Operator Bundle](https://quay.io/repository/kuadrant/authorino-operato-bundle),
+[Authorino Operator image](https://quay.io/repository/kuadrant/authorino-operator) and regarding its [manifests](https://github.com/Kuadrant/authorino-operator/blob/main/config/crd/bases/operator.authorino.kuadrant.io_authorinos.yaml)
```
guicassolato, line 62:
```suggestion
[Authorino Operator catalog image](https://quay.io/repository/kuadrant/authorino-operator-catalog) and regarding its [manifests](https://github.com/Kuadrant/authorino-operator/blob/main/config/crd/bases/operator.authorino.kuadrant.io_authorinos.yaml)
```

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,326 @@
...
+1. Pick a stable (released) version “v0.X.Y” of the operand known to be compatible with operator’s image for “v0.W.Z”; if needed, make a release of the operand first.
+2. Run the [Github Action Release operator](https://github.com/Kuadrant/authorino-operator/actions/workflows/release.yaml);
+make sure to fill all the fields:
+  * Branch containing the release workflow file – default: ‘main’
+  * Commit SHA or branch name of the operator to release – usually: ‘main’
+  * Operator version to release (without prefix) – i.e. ‘0.W.Z’
+  * Authorino version the operator enables installations of (without prefix) – i.e. ‘0.X.Y’
+  * If the release is a prerelease
+  * Bundle and catalog channels (comma-separated) – usually: 'stable'
+  * Run the GHA ‘Build and push images’ for the “v0.W.Z” tag, specifying ‘Authorino version’ equals to “0.X.Y” (without the leading “v”). This will cause the new images (bundle and catalog included) to be built and pushed to the corresponding repos in quay.io/kuadrant.
```
guicassolato, line 77:
```suggestion
3. Run the GHA ‘Build and push images’ for the “v0.W.Z” tag, specifying ‘Authorino version’ equals to “0.X.Y” (without the leading “v”). This will cause the new images (bundle and catalog included) to be built and pushed to the corresponding repos in quay.io/kuadrant.
```

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,326 @@
...
+
```
guicassolato, line 78: Missing something here about publishing the manifest bundles to the community operator hubs. That is also part of releasing, no?

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,326 @@
...
+ ```
+
+### Limitador Operator
+This operator is also implemented using the [Operator SDK](https://sdk.operatorframework.io/) and
+[Kustomize](https://kustomize.io/).
+
+#### Artifacts
+The deliverable artifacts are the [Limitador Operator image](https://quay.io/repository/kuadrant/limitador-operator),
+[Limitador Operator Bundle](https://quay.io/repository/kuadrant/limitador-operato-bundle),
+[Limitador Operator image](https://quay.io/repository/kuadrant/limitador-operator) and regarding its
```
guicassolato, line 180:
```suggestion
[Limitador Operator catalog image](https://quay.io/repository/kuadrant/limitador-operator-catalog) and regarding its
```

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,832 @@
...
+release the component. This file could be a YAML, TOML or simply a key-value file. The file should contain the following
+information:
+
+* Component/s name/s and version/s (In the case of Limitador, it should contain the name and version of the crate and the name of the server).
+* Dependencies name/s and version/s.
+* Replace version (in the case of the Operators, this version is needed for the OLM upgrade process).
+
+This file should be updated every time a new release is made, and it should be committed to the repository to the release
+branch. This file will be used by the CI/CD tool to build and release the component. In the case of the main branch, the
+file will include the next version to be released, following the floating suffix `-dev` for its version and its dependencies.
```
alexsnaps, line 334: I like the `-dev` (or some equivalent suffix) for "in progress" work... another option is to use the git SHA as [a suffix](https://www.reddit.com/r/devops/comments/ax84t0/comment/ehs2rrz/) - tho I'd prefix it with something stable so semantic versioning (as pointed in the comments of that thread). Or put the sha in the [metadata](https://semver.org/#spec-item-10), as:

>Build metadata MUST be ignored when determining version precedence.

That's if we wanted to move keep some sort of "floating release" while achieving proper tracability

So using -dev, also would prevent us from using `-alpha` or `-beta` ... or anything "lower than `-dev`, as per [semver](https://semver.org/#spec-item-11)

>Identifiers with letters or hyphens are compared lexically in ASCII sort order.

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
++version = "1.3.0"
+diff --git a/limitador/Cargo.toml b/limitador/Cargo.toml
+index 3aebf9d..d17b92b 100644
+--- a/limitador/Cargo.toml
++++ b/limitador/Cargo.toml
+@@ -1,6 +1,6 @@
+ [package]
+ name = "limitador"
+-version = "0.5.0-dev"
++version = "0.5.0"
```
guicassolato, line 124:
```suggestion
-version = "0.6.0-dev"
+version = "0.6.0"
```
?

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
+ ```diff
+diff --git a/limitador/Cargo.toml b/limitador/Cargo.toml
+index 3aebf9d..d17b92b 100644
+--- a/limitador/Cargo.toml
++++ b/limitador/Cargo.toml
+@@ -1,6 +1,6 @@
+ [package]
+ name = "limitador"
+-version = "0.5.0-dev"
++version = "0.5.0"
```
guicassolato, line 455:
```suggestion
-version = "0.6.0-dev"
+version = "0.6.0"
```

?

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
+2. Make sure your `Cargo.toml` is reflecting the proper version, see above
+3. Push the branch to remote, which should create a matching release to quay.io with the tag name based of your branch, i.e. in this case `v1.3.0`
+4. Create a tag with the `server-v` prefix, e.g. `server-v1.3.0`
+5. Push the tag to Github
+6. Create the release and release notes on [Github](https://github.com/Kuadrant/limitador/releases/new) using the tag from above, named: `vM.m.d`
+7. Delete the branch, only keep the tag used for the release
+
+##### After the release
+
+1. Create a `next` branch off `main`
```
guicassolato, line 156: Is the new branch really to be called `next` or do you mean `[next-version]-dev`?

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
+2. Update the _both_ Cargo.toml to point to the next `-dev` release
```
guicassolato, line 157:
```suggestion
2. Update _both_ Cargo.toml files to point to the next `-dev` release
```

rfcs/0000-kuadrant-ci-cd.md
guicassolato: Hmm. I see. This step is only to update the versions mentioned in the Cargo.toml files, in the `main` branch, to be the "next" one to release. The actual name of the branch does not matter.

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
+2. Run the action.
+This will commit the changes to the release branch with message "[release] Releasing Limitador crate 0.6.0 and Server 1.4.0".
+It will verify the manifests, build the images and publish them to the container registry. It also will draft a release
+on Github with the information provided by the user.
+
+3. After the release workflow is done:
+   * approve the Github draft release and publish it.
+   * change the draft PR on OperatorHub to ready to review.
+
+The following steps can be automated by the release workflow, but they could be done manually as well:
```
guicassolato, line 759:
```suggestion
The following steps can be automated by the release workflow, but they could be done manually as well:

```

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
+2. Run the action.
+This will commit the changes to the release branch with message "[release] Releasing Limitador Operator 0.8.0".
+It will verify the manifests, build the images and publish them to the container registry. It also will draft a release
+on Github with the information provided by the user.
+
+3. After the release workflow is done:
+   * approve the Github draft release and publish it.
+   * change the draft PR on OperatorHub to ready to review.
+
+The following steps can be automated by the release workflow, but they could be done manually as well:
```
guicassolato, line 784:
```suggestion
The following steps can be automated by the release workflow, but they could be done manually as well:

```

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,835 @@
...
+2. Run the GHA ‘Build and push images’ for the “v0.X.Y” tag. This will cause a new image to be built and pushed to quay.io/kuadrant/authorino.
+
+3. Create the release and release notes on [Github](https://github.com/Kuadrant/authorino/releases/new) using the tag from above, named: `Authorino vM.m.d`
+
+Notes:
+PRs merged to the main branch of Authorino cause a new image to be built (GH Action) and pushed automatically to
+quay.io/kuadrant/authorino:<git-ref> – the quay.io/kuadrant/authorino:latest tag is also moved to match the latest <git-ref>.
+
+### Authorino Operator
+The Authorino operator is similar to its operand, it fully follows the [Operator Pattern](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/)
```
guicassolato, line 63: Authorino Operator uses nowadays a mix of all options below (1, 2 and 3).

There's is a `build.yaml` file (option 1) that is created by a script/make target (option 2), and committed to each release branch.

The file contains the version of the operand the operator is bound to (in the form of a container image ref) and the value of the `replaces` directive to be used in the operator bundle manifests. I.e. all info it needs to prepare a new release.

E.g. https://github.com/Kuadrant/authorino-operator/blob/release-v0.10.0/build.yaml

The process of creating the release branch with its build file, generating the operator bundle, and publishing the GitHub release notes is triggered in GitHub, by executing the _Release operator_ action (option 3).

guicassolato:
A couple things regarding this that I'm still trying to wrap my head around are:
1. The difference between releasing Kuadrant (the suite) vs. releasing each individual component
2. The fact that we have different kinds of components, with different required steps to accomplish what we perceive as a "release", but yet discussing options that would allegedly unify it all into a single process

### 1. Releasing Kuadrant vs Releasing components of Kuadrant

This RFC is clear about its aim of implementing a "new release process for Kuadrant components", not for Kuadrant itself. I wonder if this is because somehow it's believed that releasing Kuadrant is a process that reduces simply to releasing each of its components and then that it happens naturally, perhaps with "Kuadrant" here represented by one of the components, i.e., the Kuadrant Operator, rather than the composition. Or maybe because indeed covering this other part of the process, of what characterises releasing Kuadrant itself (not the components), is beyond scope for now.

Either way, it feels like we're trying to simplify releasing Kuadrant by removing the differences involved in releasing each of its inner parts. This not only neglects the part of defining what releasing Kuadrant even means (where perhaps some of the complexity lies), but may also put a bit too much hope IMO on unifying the processes for individual and essentially different components, which leads me to my second point.

### 2. Unified process for different components

If having a single unified process that works for all components is envisioned and described as simply as having a big "Release" button for each component, that abstracts the complexities of each individual process, then I'm totally fine with it. If, on the other hand, we want to slice that into finer details, then we have to remember that releasing component X is not the same as releasing component Y.

If such differences exist between every single pair of components whose release process we compare, then we're likely better with each component owning its own release process.

It's good that we document and even automate the steps as much as we can of course, but not everything has to come top-down as "this is the way to release", nor unified we need a central release control plane which can end up turning into a bunch of if's and else's.

What I understand has become a PITA and therefore deserves some attention is that we currently have multiple ways of releasing _operators_, and all our operators are based on the same tech (Golang, Operator SDK), with artifacts published to the same hubs. Unifying how we build, publish and release operators is a good thing, IMO.

<br/>

In general, my recommendations for improving how we do releases:
1. All operators should be released the same way if possible - i.e. same scripts and tools, same types of artifacts, same publishing platforms/hubs
2. Different components should own their specific release particularities. E.g.: if releasing Authorino means creating a git tag and pushing a container image to quay.io, but Limitador requires a release branch and an additional step of publishing the crate to a particular registry, then these are details that belong to each of these components. The components should document their own release processes, and maybe Kuadrant does not have to provide much more than guidelines such as which architectures to build for, which registries to publish at, etc.
3. Define what releasing Kuadrant means in terms of the marginal effort involved, after each component has been released – from picking the stable/released versions of each component the next of Kuadrant will be based upon, to things like integration/conformance tests (if any), docs, change logs, cadence, and announcements.

IOW, this RFC could be just a set of guidelines on how to release operators and minimum requirements for the other (more special) components. In the future, describe what releasing Kuadrant means in a way that communicates to users of Kuadrant more than to maintainers what to expect.

didierofrivia:
Thanks @guicassolato for the feedback! In simple words, I understand your concerns and agree with most of your recommendations and thoughts. I'll try to expand your main 2 points with my thoughts, 

**1. Releasing Kuadrant vs Kuadrant components:**

From just the technical POV, we could define the release of Kuadrant by the process of releasing it's components, from services like Authorino and Limitador and their respective Operators to the Kuadrant Operator itself... Then, this RFC would represent the effort of solving the overhead we have when not only releasing, but providing the artifacts for different use cases. Just the unified interface/process/button abstracting or hiding the particularities of each component.

Then, regarding the "event" of releasing Kuadrant, as a project, where more aspects need to be considered, such as cadence, documentation, communication, etc.. Maybe we should address all that in a different RFC where we could be more focused. There's already something in the works in [Paul's repo](https://github.com/pmccarthy/architecture/blob/release-cadence-proposal/rfcs/0007-project-release-cadence.md).

I would say that the current title of this RFC is a bit misleading and doesn't represent well its true meaning. Maybe changing it to something like "_Kuadrant components CI/CD artifact delivery and technical release_" might be more on point.

**2. Unified process for different components**

I've briefly commented in the previous point about unifying to a single process, but it's more about establishing the process, but then each component will manage its own particularities. The operators are easier to adopt the same process, but Limitador and Authorino given their different technologies are not. This doesn't mean that the way we tag, branch, and the interface for communicating our meaning for building the artifacts can't be the same(ish).

That being said, maybe we could agree on this RFC being the technical release aspect and we still need to define the broader aspect of the _event_ of releasing.

pehala:
> That being said, maybe we could agree on this RFC being the technical release aspect and we still need to define the broader aspect of the _event_ of releasing.

In that case, shouldn't the non-technical RFC take precedence and be agreed upon before we do this one? To define what we want to do and only after that how we do it?

pehala:
> Maybe I could submit the other RFC needed and work in parallel?

I think this would be the best approach, I would also rescope this RFC to only care about Kuadrant CI/CD

alexsnaps:
Added the "Final Comment Period" to this RFC, so please try to consolidate responses within the coming week... 
Other than major disagreement, we shall merge this after this period.

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,837 @@
...
+
+## New process
+The new process ideally should be as simple as possible, and should be able to be triggered by any person, process and
+set of events (particular commits, tags, merge, etc) that needs the artifacts. It also should be able to be repeated
+in a consistent way and without the need of any human setting variables or parameters at the time of triggering it
+(at least for the most common cases).
+
+Given our current process, we can identify some similarities and particularities between the different components:
+
+![](0000-kuadrant-cicd-release/kuadrant-components.png)
```
pehala, line 310: I think these components are outdated, both MGC and DNS operator are missing

rfcs/0000-kuadrant-ci-cd.md
```diff
@@ -0,0 +1,837 @@
...
+code in different repositories. It also involves a convolution of different tools and services that are not easy to
+maintain and that are not very well integrated with each other.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Current process
+
+<details>
+<summary>Summary of the current process</summary>
```
pehala, line 29: I am missing DNS operator and MGC for both current and new process

rfcs/0000-kuadrant-ci-cd.md
mikenairn: This was on my list of things to update once the unified kuadrant PRs started merging. I can update and create a PR when i get a bit of time.

didierofrivia: @mikenairn That'd be great! thanks 🙏🏼

david-martin:
Having a split between what to release and how to release sounds reasonable.

My understanding is is that https://github.com/Kuadrant/architecture/blob/main/rfcs/0008-kuadrant-release-process.md is the 'what to release' part.
And this PR is the 'how to release' part.

I believe the only thing holding up this PR is adding the DNS operator and MGC pieces.
@mikenairn if you have no objections, this PR can be merged to main and the DNS & MGC pieces be added after?
There are a couple of other release technical details I'd like to start proposing (for kaudrantctl and dashboards), so having this rfc back on main would be helpful to propose that.

david-martin:
@didierofrivia is the rfc number of 0000 intentional, or a placeholder (I think the next number at this time is 0009)

didierofrivia:
@david-martin it's just a placeholder and will be changed once the PR it's approved, in order to avoid overlapping with other RFCs in development.

rfcs/0000-kuadrant-ci-cd.md
mikenairn: I added (not yet merged) a release doc to the dns operator repo https://github.com/Kuadrant/dns-operator/blob/88185359c731d41d8909c1ae45bf0a6ec14d0031/docs/RELEASE.md

Similar to what Gui describes here https://github.com/Kuadrant/architecture/pull/41/files#r1444455110, it has a file generated by a `prepare-release` tsk that is saved to the release branch, but its just a [Makefile](https://github.com/Kuadrant/dns-operator/blob/release-0.1/make/release.mk) overriding the appropriate variables so that all the make targets function as expected if you were to run them on that branch.

Do we want all this copied into this doc, or can we just link to that file?

re. MGC, not sure what we did for a release there tbh, but it's an operator like most others and we probably just manually did it all. Branching tagging would be the same as what is described in the dns-operator doc.

didierofrivia: Yes, that should be added here too, since we're gonna be using this document's section to gather every detail of the current release process of the components. Thanks!

maleck13:
This is coming up on a year old. Is it stale should we close or update?

didierofrivia:
@maleck13 So far this is being consulted when necessary, because of the current process sectionÑ https://github.com/Kuadrant/architecture/pull/41/files#diff-eeceecbdf5c1a6fc2cc490b55dd4351651d64c8a740458325511493e52f7616aR29 . However it's a bit out of date, I'll try to find some time to update it in the coming weeks

PR #46, opened 2024-01-11
Kuadrant Release Process

didierofrivia:
Partially closes https://github.com/Kuadrant/architecture/issues/59

Technical aspects of the process: https://github.com/Kuadrant/architecture/pull/41

### DIFF
--- a/rfcs/0000-kuadrant-release-process.md
+++ b/rfcs/0000-kuadrant-release-process.md
@@ -9,7 +9,8 @@
 [summary]: #summary
 
 Kuadrant is a set of components that their artifacts are built and delivered independently. This RFC aims to define every
-aspect of the event of releasing a new version of the whole, in terms of versioning, cadence, communication, channels, etc.
+aspect of the event of releasing a new version of the whole, in terms of versioning, cadence, communication, channels,
+handover to other teams, etc.
 
 # Motivation
 [motivation]: #motivation
@@ -22,14 +23,11 @@ in charge of the release process, which is not ideal.
 [guide-level-explanation]: #guide-level-explanation
 
 First, we need to define what releasing Kuadrant means, in a clear and transparent way that communicates to the community
-what's happening and what to expect. This includes:
-
-- Components included in each release
-- Versioning
-- Cadence
-- Repositories and Hubs
-- Documentation
-- Communication
+what's happening and what to expect. The Kuadrant suite is composed of several components, each of them with its own
+set of artifacts and versioning scheme. Defining the release process of the whole suite is a complex task, and it's
+not only about the technical details of releasing the components, but also about the communication and transparency
+with the community, the definition of the frequency of the releases, and when it's ready to be handover to other teams like
+QA. This section aims to provide guidelines for the different aspects of the release process.
 
 ## Components and versioning
 
@@ -43,24 +41,53 @@ The set of components that are part of the _Kuadrant suite_ are the following:
 - [Multicluster Gateway Controller](https://github.com/Kuadrant/multicluster-gateway-controller): Provides multi-cluster
 connectivity and global load balancing.
 - [Kuadrant Operator](https://github.com/Kuadrant/kuadrant-operator/): The Operator to install and manage the lifecycle
-of the Kuadrant components deployments.
+  of the Kuadrant components deployments.
 
-Each of them needs to be versioned independently, and the versioning scheme should follow [Semantic Versioning](https://semver.org/).
-The technical details of how to release each component are out of the scope of this RFC and could be found in the
-[Kuadrant components CI/CD](https://github.com/Kuadrant/architecture/pull/41) RFC.
+Each of them needs to be versioned independently, and the versioning scheme should follow [Semantic Versioning]. At
+the time of cutting a release for any of them, it's important to keep in mind what section of the version to bump,
+given a version number MAJOR.MINOR.PATCH, increment the:
+
+* MAJOR version when you make incompatible API changes
+* MINOR version when you add functionality in a backward compatible manner
+* PATCH version when you make backward compatible bug fixes
+
+Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
+
+A more detailed explanation of the versioning scheme can be found in the [Semantic Versioning](https://semver.org/) website.
 
 By releasing a new version of Kuadrant, we mean releasing a new version of every component in the suite, with their
 corresponding semantic versioning, and being the version of the **Kuadrant Operator** the one that defines the version of
 the whole suite.
 
+The technical details of how to release each component are out of the scope of this RFC and could be found in the
+[Kuadrant components CI/CD](https://github.com/Kuadrant/architecture/pull/41) RFC.
+
 ## Cadence
 
-The frequency of the previously mentioned components, with the exception of the **Kuadrant Operator**, could vary
-depending on the particular component needs. However, the **Kuadrant Operator** should be released every 2 weeks,
-including the latest released version of every component in the suite.
+Once the project is stable enough, and it's adoption increases, the community will be expecting a certain degree of
+commitment from the maintainers, and that includes a regular release cadence. The frequency of the releases of the
+different components could vary depending on the particular component needs. However, the **Kuadrant Operator**
+it's been discussed in the past that it should be released every 2 weeks, including the latest released version
+of every component in the suite. There's another RFC that focuses on the actual frequency of each component, one could
+refer to the [Kuadrant Release Cadence RFC](https://github.com/pmccarthy/architecture/blob/release-cadence-proposal/rfcs/0007-project-release-cadence.md).
 
-For a more in-depth analysis of the cadence of each component, please refer to the
-[Kuadrant Release Cadence RFC](https://github.com/pmccarthy/architecture/blob/release-cadence-proposal/rfcs/0007-project-release-cadence.md).
+There are a few reasons for this:
+
+- Delivering Unparalleled Value to Customers: Regular releases can provide customers with regular updates and improvements.
+  These updates can include new features and essential bug fixes, thus enhancing the overall value delivered to the customers.
+- Maximizing Deployment Efficiency: By releasing software at regular intervals, teams can align their activities with
+  available resources and environments, ensuring optimal utilization. This leads to increased efficiency in the deployment process.
+- Risk Management: Regular releases can help identify and fix issues early, reducing the risk of major failures that could
+  affect customers.
+- Feedback Cycle: Regular releases allow for quicker feedback cycles. This means that any issues or improvements
+  identified by users can be addressed promptly, leading to a more refined product over time.
+- Synchronization: Regular releases can help synchronize work across different teams or departments, creating a more
+  reliable, dependable solution development and delivery process.
+- Reduced Complexity: Managing a smaller number of releases can reduce complexity. For example, having many different
+  releases out in the field can lead to confusion and management overhead.
+
+By committing to a release cadence, software projects can benefit from improved efficiency, risk management, faster feedback
+cycles, synchronization, and reduced complexity.
 
 ## Repositories and Hubs
 
@@ -109,52 +136,6 @@ other teams within the organization. A few examples of the communication channel
 - Slack channel, in Red Hat and Kubernetes workspaces
 - Blog post, if applicable
 - Social media, if applicable
-
-# Reference-level explanation
-[reference-level-explanation]: #reference-level-explanation
-
-As mentioned in the previous section, the Kuadrant suite is composed of several components, each of them with its own
-set of artifacts and versioning scheme. Defining the release process of the whole suite is a complex task, and it's
-not only about the technical details of releasing the components, but also about the communication and transparency
-with the community. It also involves the definition of the cadence of the releases, and the frequency of the releases
-of each component. This section aims to provide a more in-depth analysis of the different aspects of the release process.
-
-## Components and versioning
-
-Taking into account the different components listed in the [Guide-level explanation section](#guide-level-explanation),
-at the time of cutting a release for any of them, it's important to keep in mind what section of the version to bump,
-given a version number MAJOR.MINOR.PATCH, increment the:
-
-* MAJOR version when you make incompatible API changes
-* MINOR version when you add functionality in a backward compatible manner
-* PATCH version when you make backward compatible bug fixes
-
-Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
-
-A more detailed explanation of the versioning scheme can be found in the [Semantic Versioning](https://semver.org/) website.
-
-## Cadence
-
-Once the project is stable enough, and it's adoption increases, the community will be expecting a certain degree of
-commitment from the maintainers, and that includes a regular release cadence. The frequency of the releases of the
-different components could vary depending on the particular component needs. However, the **Kuadrant Operator** as mentioned
-in the previous section, should be released every 2 weeks as it's been already proposed, including the latest released
-version of every component in the suite. There are a few reasons for this:
-
-- Delivering Unparalleled Value to Customers: Regular releases can provide customers with regular updates and improvements.
-These updates can include new features and essential bug fixes, thus enhancing the overall value delivered to the customers.
-- Maximizing Deployment Efficiency: By releasing software at regular intervals, teams can align their activities with
-available resources and environments, ensuring optimal utilization. This leads to increased efficiency in the deployment process.
-- Risk Management: Regular releases can help identify and fix issues early, reducing the risk of major failures that could
-affect customers.
-- Feedback Cycle: Regular releases allow for quicker feedback cycles. This means that any issues or improvements
-identified by users can be addressed promptly, leading to a more refined product over time.
-- Synchronization: Regular releases can help synchronize work across different teams or departments, creating a more
-reliable, dependable solution development and delivery process.
-- Reduced Complexity: Managing a smaller number of releases can reduce complexity. For example, having many different
-releases out in the field can lead to confusion and management overhead.
-
-By committing to a release cadence, software projects can benefit from improved efficiency, risk management, faster feedback cycles, synchronization, and reduced complexity.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/rfcs/0000-kuadrant-release-process.md
+++ b/rfcs/0000-kuadrant-release-process.md
@@ -104,8 +104,9 @@ There are a few reasons for this:
 
 - Delivering Unparalleled Value to Users: Regular releases can provide users with regular updates and improvements.
   These updates can include new features and essential bug fixes, thus enhancing the overall value delivered to the users.
-- Maximizing Deployment Efficiency: By releasing software at regular intervals, teams can align their activities with
-  available resources and environments, ensuring optimal utilization. This leads to increased efficiency in the deployment process.
+- Resource Optimization: By releasing software at regular intervals, teams can align their activities with
+  available resources and environments, ensuring optimal utilization. This leads to increased efficiency in the deployment
+  process and reduces the risk of resource wastage.
 - Risk Management: Regular releases can help identify and fix issues early, reducing the risk of major failures that could
   affect users.
 - Feedback Cycle: Regular releases allow for quicker feedback cycles. This means that any issues or improvements
@@ -163,7 +164,6 @@ process of updating the documentation is simple and consists of the following st
 1. Update the documentation in the corresponding component repository.
 2. Follow the instruction in [https://github.com/Kuadrant/docs.kuadrant.io/](https://github.com/Kuadrant/docs.kuadrant.io/)
 to update the Docs pointers to the tag or branch of the component repository that contains the updated documentation.
-file, to point to the tag or branch of the component repository that contains the updated documentation.
 3. Once the changes are merged to main, the workflow that updates the website will be triggered, and the documentation
 will be updated.
 4. If for some reason it's needed to trigger the workflow manually, one can do it from the GitHub Actions tab in the
--- a/rfcs/0000-kuadrant-release-process.md
+++ b/rfcs/0000-kuadrant-release-process.md
@@ -176,44 +176,30 @@ other teams within the organization. A few examples of the communication channel
 - Blog post, if applicable
 - Social media, if applicable
 
-# Drawbacks
-[drawbacks]: #drawbacks
-
-Why should we *not* do this?
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-- Why is this design the best in the space of possible designs?
-- What other designs have been considered and what is the rationale for not choosing them?
-- What is the impact of not doing this?
+The alternative to the proposal is to keep the current process, which is not ideal and leads to confusion and lack of
+transparency.
 
 # Prior art
 [prior-art]: #prior-art
 
-Discuss prior art, both the good and the bad, in relation to this proposal.
-A few examples of what this can include are:
+There's been an organically grown process for releasing new versions of the Kuadrant suite, which is not documented and
+it's been changing over time. However, there are some documentation for some of the components, worth mentioning:
 
-- Does another project have a similar feature?
-- What can be learned from it? What's good? What's less optimal?
-- Papers: Are there any published papers or great posts that discuss this? If you have some relevant papers to refer to, this can serve as a more detailed theoretical background.
-
-This section is intended to encourage you as an author to think about the lessons from other tentatives - successful or not, provide readers of your RFC with a fuller picture.
-
-Note that while precedent set by other projects is some motivation, it does not on its own motivate an RFC.
+* [Authorino release process](https://docs.google.com/document/d/1tLveyv8Zwe0wKyfUTWOlEnFeMB5aVGqIVDUjVYWax0U)
+* [Limitador release process](https://github.com/Kuadrant/limitador/blob/main/RELEASE.md)
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-- What parts of the design do you expect to resolve through the RFC process before this gets merged?
-- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
-- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+* What would be Kuadrant support policy?
+* How many version are we going to back-port security and bug fixes to?
+* What other teams need to be involved in the release process?
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
-Think about what the natural extension and evolution of your proposal would be and how it would affect the platform and project as a whole. Try to use this section as a tool to further consider all possible interactions with the project and its components in your proposal. Also consider how this all fits into the roadmap for the project and of the relevant sub-team.
-
-This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related.
-
-Note that having something written down in the future-possibilities section is not a reason to accept the current or a future RFC; such notes should be in the section on motivation or rationale in this or subsequent RFCs. The section merely provides additional information.
+Once the release process is accepted and battle-tested, we could aim to automate the process as much as possible.
--- a/rfcs/0000-kuadrant-release-process.md
+++ b/rfcs/0000-kuadrant-release-process.md
@@ -45,8 +45,8 @@ environments.
 - [Kuadrant Operator](https://github.com/Kuadrant/kuadrant-operator/): The Operator to install and manage the lifecycle
   of the Kuadrant components deployments.
 
-Each of them needs to be versioned independently, and the versioning scheme should follow [Semantic Versioning]. At
-the time of cutting a release for any of them, it's important to keep in mind what section of the version to bump,
+Each of them needs to be versioned independently, and the versioning scheme should follow [Semantic Versioning](https://semver.org/).
+At the time of cutting a release for any of them, it's important to keep in mind what section of the version to bump,
 given a version number MAJOR.MINOR.PATCH, increment the:
 
 * MAJOR version when you make incompatible API changes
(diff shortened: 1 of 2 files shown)


rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+are ready for testing. This notification can be done through comments on GitHub issues or via other communication channels
+like Slack.
+5. Testing: The QA team tests the release candidate, checking for any bugs or issues. They can use GitHub's issue tracking
+system to report any problems they find.
+6. Iterate: Based on the feedback from the QA team, the development team makes any necessary adjustments and repeats the
+process until the release candidate is deemed ready for production.
+7. Publish Release: Once the release candidate is ready, the engineering team publishes the release. This means that the
+Github release draft is published, and the artifacts are uploaded to the corresponding registries.
+
+## Cadence
```
Boomatang, line 92: This question might not belong here but looking through the links I was not able to find the answer or the correct place for the question. 

If we are releasing the Kuadrant Operator every two weeks, what is our support policy? How many version are we going to back port security and bug fixes to?

didierofrivia: It's a valid question, and important one to have it registered in the RFC. I'd say that we should include not only engineering and QA, but what would be the expectation from Product as well.

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+
+Once the project is stable enough, and it's adoption increases, the community will be expecting a certain degree of
+commitment from the maintainers, and that includes a regular release cadence. The frequency of the releases of the
+different components could vary depending on the particular component needs. However, the **Kuadrant Operator**
+it's been discussed in the past that it should be released every 2 weeks, including the latest released version
```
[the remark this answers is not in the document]
didierofrivia: Will reword it was you mention

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+- Feature Name: `kuadrant-release-process`
+- Start Date: 2024-01-11
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Kuadrant is a set of components that their artifacts are built and delivered independently. This RFC aims to define every
+aspect of the event of releasing a new version of the whole, in terms of versioning, cadence, communication, channels,
```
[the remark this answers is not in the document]
didierofrivia: I reckon that Paul's one will address the actual definition of cadence. Say, for some components, for the whole, etc. Kind of agreeing on the frequency. 
And in this one, the importance of it. Does it make sense?

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+
+* MAJOR version when you make incompatible API changes
+* MINOR version when you add functionality in a backward compatible manner
+* PATCH version when you make backward compatible bug fixes
+
+Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
+
+A more detailed explanation of the versioning scheme can be found in the [Semantic Versioning](https://semver.org/) website.
+
+By releasing a new version of Kuadrant, we mean releasing a new version of every component in the suite, with their
```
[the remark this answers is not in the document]
didierofrivia: Will reword it stating that some components might still be the same as the previous release.

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+
+- Feature Name: `kuadrant-release-process`
+- Start Date: 2024-01-11
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Kuadrant is a set of components that their artifacts are built and delivered independently. This RFC aims to define every
```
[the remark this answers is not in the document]
didierofrivia: I think that the difference is Kuadrant operator (the orchestrator of the suite) and Kuadrant components (Authorino, Limitador, MGC, etc). I'm saying this because I thought that QA mainly checks the versions included in the latest released version of Kuadrant Operator. I could re write it as per each component... which kind of will remove the responsibility of deciding the cadence and leave it to Pauls RFC (?)

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+* MAJOR version when you make incompatible API changes
+* MINOR version when you add functionality in a backward compatible manner
+* PATCH version when you make backward compatible bug fixes
+
+Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
+
+A more detailed explanation of the versioning scheme can be found in the [Semantic Versioning](https://semver.org/) website.
+
+By releasing a new version of Kuadrant, we mean releasing a new version of every component in the suite, with their
+corresponding semantic versioning, and being the version of the **Kuadrant Operator** the one that defines the version of
```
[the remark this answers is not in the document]
didierofrivia: Sounds like a good candidate for the community call indeed. It's something I kind of been suggesting, or at least a given xD

rfcs/0000-kuadrant-release-process.md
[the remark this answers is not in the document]
didierofrivia: On another note, it mentions that the artifacts from each component are build (and delivered) independently, but might not be subject to a new event of releasing.

rfcs/0000-kuadrant-release-process.md
didierofrivia: I will add it to the "Unresolved questions" section

rfcs/0000-kuadrant-release-process.md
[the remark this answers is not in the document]
guicassolato: > actual release process should be the same for every component

I see this differently.

IMO, it does not _have to_ be, neither it _can_ always be the same process for every component. Rather (as originally phrased), the components are independent from the whole and can be released independently, following their own particular release processes, respecting each one its own specificity. It's not top-down; it's bottom-up.

For some component, it may make sense, for example, doing release branches, while for others it's not needed and just a tag will do. Some components may involve publishing artifacts to a particular repo, while others do not. Because of these and so many other differences, each component is a project and has its own history. Thinking the opposite risks creating coupling between the components and generates scalability issues to the organisation.

Of course, we don't want multiple ways to do the exact same thing. To reduce overhead, if 2 components share a common step within their corresponding release processes, we should try performing that step in the same way – by resusing tools, automation, docs, etc across components. (Although, I think this would be something that belongs to #41 and not to this PR.)

Correct if I'm wrong, but I understand this RFC is for how to lock a version of what we call "Kuadrant", which is not the sum of its parts, but has an identity of its own. Therefore, it's not to force components all to stick to a same process but to highlight precisely this difference between the sum of the parts and the whole. Maybe it's a description of the steps to follow _after_ all components have been released, and perhaps some general guidelines that all components need to comply with, otherwise releasing Kuadrant would be impossible.

If there's a components that confuses itself with the whole, then an explanation of that and some words about how releasing that one component wraps up a release of Kuadrant itself maybe (if that's the case.)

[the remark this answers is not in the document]
pehala: I would say this RFC should be high-level and only mention the release process, not the actual steps, e.g. how we build those artefacts should not be part of this (That is part of the Kuadrant CI/CD RFC and that indeed differs). I do not see a difference between component release and Kuadrant release in a high-level process. For each component there should be an RC build, and for each component QA should happen, exactly the way described here. The doc is also present for all components on the `kuadrant.io`, communication channels are also the same

So basically in my eyes, the process is the key section and it should apply to all component releases and the entire Kuadrant release at the same time.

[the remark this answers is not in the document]
guicassolato: @pehala, I think you may be seeing Kuadrant == MGC (or any other components that currently sits at the root of the tree of dependencies.)

In case my view wasn't clear... IMO, Kuadrant is a composition of things (Kuadrant Operator, Authorino, Limitador, MGC) and a way to use those things. Let's call this second part "other stuff" – docs, templates, maybe an installer, and, in general, other files we can/could provide in the "Kuadrant package" that are not part of the none of the individual components per se.

So:

```
K = KOp + A + L + MGC + stuff
```

A given version `K_v1` of Kuadrant can be, for example, a composition of Kuadrant Operator v0.3, Authorino v15, Limitador v0.8 and MGC v0.1. I.e.:

```
K_v1 = KOp_v0.3 + A_v15 + L_v0.8 + MGC_v0.1 + stuff
```

If we want to bump the version of, say, Limitador that is used in/by Kuadrant to v0.9 - not because Limitador v0.9 has just been released; maybe the latest Limitador is already v2.5 by the time this happens –, then the Kuadrant composition will change. IOW, Kuadrant will be a different thing than what it was at v1. Maybe we are talking about Kuadrant v2:

```
K_v2 = KOp_v0.3 + A_v15 + L_v0.9 + MGC_v0.1 + stuff
```

The only thing that changed was Limitador and, maybe, some of the other `stuff`, like new docs saying that now Kudrant v2 comes with Limitador v0.9.

Each component, as well as the suite itself, has (should have) its own set of artifacts, its own version number, its own change log and release notes, its own testing and release process. Kuadrant is no longer equivalent to the Kuadrant Operator. MGC may have taken its place as the new root, but we need to see (...and test, ...and version, ...and release) Kuadrant independently from it as well.

This is the way I see Kuadrant, at least. I may change my mind if most people agree that Kuadrant == MGC, in which case I would argue for changes beyond the defining our release process, starting by renaming those repos, MGC and Kuadrant Operator, and moving some code and responsibilities around. In the end, it would all contribute to defining what releasing Kuadrant means and how it should look like.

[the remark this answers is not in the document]
didierofrivia: I kind of like the idea you bring defining what would be the version of _Kuadrant_, quoting your formula:

> K = KOp + A + L + MGC + stuff

In that case, and if everyone/most people agree, I could add another section regarding the particular versioning and grouping that will define it.

alexsnaps:
Added the "Final Comment Period" to this RFC, so please try to consolidate responses within the coming week... 
Other than major disagreement, we shall merge this after this period.

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,205 @@
...
+
+- Feature Name: `kuadrant-release-process`
+- Start Date: 2024-01-11
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Kuadrant is a set of components that their artifacts are built and delivered independently. This RFC aims to define every
```
thomasmaas, line 11:
```suggestion
Kuadrant is a set of components whose artifacts are built and delivered independently. This RFC aims to define every
```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,205 @@
...
+A more detailed explanation of the versioning scheme can be found in the [Semantic Versioning](https://semver.org/) website.
+
+By releasing a new version of Kuadrant, we mean releasing the set of components with their corresponding semantic versioning,
+some of them maybe freshly released, or others still using versioning from the previous one, and being the version of the
+**Kuadrant Operator** the one that defines the version of the whole suite.
+
+The technical details of how to release each component are out of the scope of this RFC and could be found in the
+[Kuadrant components CI/CD](https://github.com/Kuadrant/architecture/pull/41) RFC.
+
+## Handover to QA
```
thomasmaas, line 65: I keep thinking that this should not be part of the release process upstream. We should make sure the test suite is always as up to date as possible and those tests that exist should pass. But that's it. Any gap that exists between features and QE coverage should be minimized by optimizing our development process, not by adding a post development QE process. By adding such a post development QE step we've already thrown in the towel.

pehala: But the reason for having a small QE part on the release is not about testsuite, I believe there are tasks which are simply unfeasible to run frequently (e.g. huge overhead) and hence are ideal for the release testing, for example, running on all supported configurations (single + multicluster).

Some tasks can only be done after we have a final bundle completed, like testing quickstart, verifying OLM deliverables, and verifying that all images are correct, which cannot be done beforehand.

thomasmaas: Ok, maybe the term "Handover to QE" is what triggered my response as it sounds heavy and slow to me. Also, do we "support configurations"? I'd say we "enable configurations". And if there are issues they can be fixed. I guess what I'm saying is that we should not drop the full weight of downstream on upstream. That doesn't mean we should ship sloppy software but it does mean we can make a different risk calculation.

pehala: I agree with both making a better name for the step and that we should not drop the full weight downstream on upstream. I believe the only thing that have not yet reached an agreement on is the risk calculation itself.

While it would be completely ok to ship without QE step at release at this stage. However, I believe the value of additional QE testing on release outweighs the delay it causes (e.g. the 2-3 days maximum that we commit ourselves to do). We are also definitely not treating it as a downstream release, we do have different requirements for upstream (only automation run, fixed time to test vs test everything etc).

I do not see the inclusion of the "QE" step in a release process as a downstream thing, but as a last sanity check that everything is ok and as a place to do testing on the all assembled pieces compared to individual components (Note: This argument does not hold for individual component releases, however, the still the last run of the automation on a final build is still a good sanity check). There is a fine balance of testing required that we need to find and what exactly should be tested for upstream but I do not think that eliminates the step, only reduces its length.

Regarding the `support configuration` vs `enable configurations`, enable is a better name for it for sure (I keep defaulting to downstream language because I have no idea how to call it otherwise), however, I think it is also an important part of the testing for a project where multicluster is a key feature, so testing it on dedicated multiple production-grade clusters makes sense, just not on every PR

thomasmaas: This is a helpful conversation and I think we're pretty close to agreeing. How about we call the step "QE sanity check"? We might also add that we aim for this step to take 0 extra time and to be fully automated at some moment in the future but that for now we agree on an expected duration of 1-2 days and a maximum duration of 3 days.

pehala: I am fully onboard with that idea! Thanks for suggesting it!

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,205 @@
...
+4. If for some reason it's needed to trigger the workflow manually, one can do it from the GitHub Actions tab in the
+[docs.kuadrant.io](https://github.com/Kuadrant/docs.kuadrant.io/actions/workflows/build.yaml) (`Actions > ci > Run Workflow`).
+
+## Communication
+
+Another important aspect of releasing a new version of the Kuadrant suite is the communication with the community and
+other teams within the organization. A few examples of the communication channels that need to be updated are:
+
+- Changelog generation
+- Release notes
```
eguzki, line 173: I like how Istio is doing this.

Release announcements

Screenshot 2024-02-16 at 14-59-30 Release Announcements


X.Y.x patch releases


Screenshot 2024-02-16 at 15-00-23 1 20 x Releases

New release notes

Screenshot 2024-02-16 at 15-00-52 Announcing Istio 1 20 3

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,205 @@
...
+
+Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
+
+A more detailed explanation of the versioning scheme can be found in the [Semantic Versioning](https://semver.org/) website.
+
+By releasing a new version of Kuadrant, we mean releasing the set of components with their corresponding semantic versioning,
+some of them maybe freshly released, or others still using versioning from the previous one, and being the version of the
+**Kuadrant Operator** the one that defines the version of the whole suite.
+
+The technical details of how to release each component are out of the scope of this RFC and could be found in the
```
eguzki, line 62: This is the most interesting aspect for me. So I leave you (capable) guys deciding the other stuff

eguzki (commented):
left some comments

rfcs/0000-kuadrant-release-process.md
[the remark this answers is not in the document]
pehala: I am all for the section defining Kuadrant and its related components. For the issue of the individual component release process, my suggestion would be to use this process for operator-based components that are released individually, meaning, this release process would be used for these components:

- Kuadrant as a whole
- Authorino + Authorino operator
- Limitador + Limitador Operator
- DNSOperator
- Multicluster Gateway Controller

These components are all operator-based and distributed the same way; thus, it makes sense for them to follow the same process.

@guicassolato @didierofrivia Is that ok with you? Does that even make sense?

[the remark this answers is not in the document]
didierofrivia: @pehala thanks for you comment! Let me elaborate and see if I understood correctly.... within a release of Kuadrant, one will encounter Operators released independently, which I reckon all of us agree (at least to a certain point) that given their similarities its easy to understand they will follow the same technical process of releasing... 

But given your last comment and trying to summarize the thread,  are we meaning to do...

```
Kuadrant vx.y.z = Kuadrant Operator vx.y.z + Authorino Operator vx.y.z + Limitador Operator vx.y.z + DNS Operator vx.y.z + MGC Controller vx.y.z
```
Where each dependency component will be released independently and grouped under a version of Kuadrant and then we could assign the Kuadrant version in the same fashion of the other components.

OR

Keeping the same way of independently releasing but matching the version of Kuadrant to its Kuadrant Operator:

```
Kuadrant vx.y.z = Kuadrant Operator vx.y.z =  Authorino Operator vx.y.z + Limitador Operator vx.y.z + DNS Operator vx.y.z + MGC Controller vx.y.z
```

Maybe I'm missing the point, but that is what we still haven't agreed on. 

@guicassolato @pehala how close are we to (dis)agree on this? Thanks in advance!

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+- [Authorino Operator](https://github.com/Kuadrant/authorino-operator): A Kubernetes Operator to manage Authorino instances.
+- [Limitador](https://github.com/Kuadrant/limitador): A generic rate-limiter written in Rust.
+- [Limitador Operator](https://github.com/Kuadrant/limitador-operator/): A Kubernetes Operator to manage Limitador deployments.
+- [Wasm Shim](https://github.com/Kuadrant/wasm-shim/): A Proxy-Wasm module written in Rust, acting as a shim between Envoy and Limitador.
+- [Multicluster Gateway Controller](https://github.com/Kuadrant/multicluster-gateway-controller): Provides multi-cluster
+connectivity and global load balancing.
+- [Kuadrant Operator](https://github.com/Kuadrant/kuadrant-operator/): The Operator to install and manage the lifecycle
+  of the Kuadrant components deployments.
+
+Each of them needs to be versioned independently, and the versioning scheme should follow [Semantic Versioning]. At
```
guicassolato, line 46:
```suggestion
Each of them needs to be versioned independently, and the versioning scheme should follow [Semantic Versioning](https://semver.org/). At
```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+The technical details of how to release each component are out of the scope of this RFC and could be found in the
+[Kuadrant components CI/CD](https://github.com/Kuadrant/architecture/pull/41) RFC.
+
+## QA Sanity Check
+
+Probably the most important and currently missing step in the release process is the green flagging from the Quality Assurance
+(QA) team. The QA team is responsible for testing the different components of the Kuadrant suite, and they need to
+be aware of the new version of the suite that is going to be released, what are the changes that are included, bug fixes
+and new features in order they can plan their testing processes accordingly. This check is not meant to be a fully fledged
+assessment from the QA team as it would be with a downstream project when it's handover to them, it's aimed to not take
```
guicassolato, line 71:
```suggestion
assessment from the QA team when it's handover to them, it's aimed to not take
```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+more than 1-2 days, and be expected to be fully automated in the future. This step will happen once the release candidate
```
guicassolato, line 72:
```suggestion
more than 1-2 days, and ideally expected to be fully automated. This step will happen once the release candidate
```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+is ready, and it's been tested by the Engineering team. The QA team should be notified in advance of the release, and
+they should be able to assert the release candidate is ready before the actual release happens.
```
guicassolato, line 74:
```suggestion
```suggestion
has no PRs pending to be merged, and it has been tested by the Engineering team. The QA team should work closely to the engineering throughout the process, both teams aiming for zero handover time and continuous delivery mindset, so immediate testing can be triggered on release candidates once handed over. This process should happen without the need of formal communication between the teams or any overhead in general, but by keeping constant synergy between quality and product engineering instead.
``````

File: rfcs/0000-kuadrant-release-process.md [hunk 15]
```diff
@@ -0,0 +1,207 @@
...
+
+There is an ideal time to hand over to the QA team for testing, especially since we are using GitHub for orchestration,
+we could briefly define it in the following steps:
+
+1. Complete Development Work: The engineering team completes their work included in the milestone.
+2. Create Release Candidate: The engineering team creates Release Candidate builds and manifests for all components required
+for the release
+3. Notify QA Team: At this time, the engineering team can notify the QA team that the deliverables for the release candidate
```
guicassolato, line 82: I personally think this step should not exist.

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+are ready for testing. This notification can be done through the GitHub Kuadrant board, or simply a public message in the
+Kuadrant Slack channel.
+4. Flagging/Testing: The QA team do the actual assertion/testing of the release candidate, checking for any obvious bugs or issues.
+Then QA reports all the bugs as GitHub issues and communicates testing status back publicly on Slack and/or email.
+5. Iterate: Based on the feedback from the QA team, the development team makes any necessary adjustments and repeats the
```
guicassolato, line 87: To be consistent with the terminology, should we call it "engineering team", "development team", "product engineering team" (since QE is also "engineering")?

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+process until the release candidate is deemed ready for production.
+6. Publish Release: Once QA communicates that the testing has been successfully finished, the engineering team will publish
+the release both on Github and in the corresponding registries, updates documentation for the new release, and communicates
+it to all channels specified in Communication section.
+
+## Cadence
+
+Once the project is stable enough, and it's adoption increases, the community will be expecting a certain degree of
```
guicassolato, line 95:
```suggestion
Once the project is stable enough, and its adoption increases, the community will be expecting a certain degree of
```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+it's been discussed in the past that it should be released every 3-4 weeks initially, including the latest released version
+of every component in the suite. There's another RFC that focuses on the actual frequency of each component, one could
+refer to the [Kuadrant Release Cadence RFC](https://github.com/pmccarthy/architecture/blob/release-cadence-proposal/rfcs/0007-project-release-cadence.md).
+
+There are a few reasons for this:
+
+- Delivering Unparalleled Value to Customers: Regular releases can provide customers with regular updates and improvements.
+  These updates can include new features and essential bug fixes, thus enhancing the overall value delivered to the customers.
+- Maximizing Deployment Efficiency: By releasing software at regular intervals, teams can align their activities with
+  available resources and environments, ensuring optimal utilization. This leads to increased efficiency in the deployment process.
```
guicassolato, line 107: Is this about cross product/project/dependency compatibility? What "available resources and environments" are we talking about? Or maybe it's not clear to me who the "teams" are in this case.

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+- Risk Management: Regular releases can help identify and fix issues early, reducing the risk of major failures that could
+  affect customers.
```
guicassolato, line 109:
```suggestion
  affect users.
```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+|                                 | limitador-operator manifests                   | [OperatorHub.io](https://operatorhub.io/operator/limitador-operator)                   |
+| Wasm Shim                       | wasm-shim images                               | [Quay.io](https://quay.io/repository/kuadrant/wasm-shim)                               |
+| Multicluster Gateway Controller | multicluster-gateway-controller images         | [Quay.io](https://quay.io/repository/kuadrant/multicluster-gateway-controller)         |
+|                                 | multicluster-gateway-controller-bundle images  | [Quay.io](https://quay.io/repository/kuadrant/multicluster-gateway-controller-bundle)  |
+|                                 | multicluster-gateway-controller-catalog images | [Quay.io](https://quay.io/repository/kuadrant/multicluster-gateway-controller-catalog) |
+| Policy Controller               | policy-controller images                       | [Quay.io](https://quay.io/repository/kuadrant/policy-controller)                       |
```
guicassolato, line 143: Policy Controller was not mentioned as one of the components before.

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+
+The documentation for the Kuadrant suite is compiled and available on the [Kuadrant website](https://kuadrant.io/). One
+can find the source of the documentation within each component repository, in the `docs` directory. However, making this
+information available on the website is a manual process, and should be done by the maintainers of the project. The
+process of updating the documentation is simple and consists of the following steps:
+
+1. Update the documentation in the corresponding component repository.
+2. In [https://github.com/Kuadrant/docs.kuadrant.io/](https://github.com/Kuadrant/docs.kuadrant.io/), update the value of
+the `import_url` within the `multirepo` in the `plugins` section of the [mkdocs.yml](https://github.com/Kuadrant/docs.kuadrant.io/blob/main/mkdocs.yml)
+file, to point to the tag or branch of the component repository that contains the updated documentation.
```
guicassolato, line 163:
```suggestion
2. Follow the instruction in [https://github.com/Kuadrant/docs.kuadrant.io/](https://github.com/Kuadrant/docs.kuadrant.io/) to update the Docs pointers to the tag or branch of the component repository that contains the updated documentation.

```

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,207 @@
...
+The alternative to the proposal is to keep the current process, which is not ideal and leads to confusion and lack of
+transparency.
+
+# Prior art
+[prior-art]: #prior-art
+
+There's been an organically grown process for releasing new versions of the Kuadrant suite, which is not documented and
+it's been changing over time. However, there are some documentation for some of the components, worth mentioning:
+
+* [Authorino release process](https://docs.google.com/document/d/1tLveyv8Zwe0wKyfUTWOlEnFeMB5aVGqIVDUjVYWax0U)
```
guicassolato, line 194: I'm afraid this document is not accessible outside of Red Hat bu default. Shall we follow Limitador and move it into the Authorino repo?

rfcs/0000-kuadrant-release-process.md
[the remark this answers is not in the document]
pehala: When it comes to this, I have no preference, so whatever works for you

rfcs/0000-kuadrant-release-process.md
didierofrivia: In this case teams would refer to (as of today) Engineering and QA, and in the future probably include Product.  Of course this makes more sense when talking about Downstream, but I think it could apply to Upstream too.

rfcs/0000-kuadrant-release-process.md
didierofrivia: Now removed, since the Kuadrand Operator will manage the DNS and TLS policies. Also added the DNS Operator

rfcs/0000-kuadrant-release-process.md
didierofrivia: Do you reckon it could be implicitly in the following step "Flagging/Testing" ?

rfcs/0000-kuadrant-release-process.md
guicassolato: I think I know what mean. I first read this from a different angle, of users who have Kuadrant in their CI pipelines.

If it's about Kuadrant teams, "deployment" seems a little out of place IMO, but maybe it's me.

Perhaps "Resource Optimization"? As in teams working together and in sync can plan their resources better (e.g. avoid idle servers, etc?)

Could it be overlapping with _Synchronization_ a little?

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,211 @@
...
+
+The documentation for the Kuadrant suite is compiled and available on the [Kuadrant website](https://kuadrant.io/). One
+can find the source of the documentation within each component repository, in the `docs` directory. However, making this
+information available on the website is a manual process, and should be done by the maintainers of the project. The
+process of updating the documentation is simple and consists of the following steps:
+
+1. Update the documentation in the corresponding component repository.
+2. Follow the instruction in [https://github.com/Kuadrant/docs.kuadrant.io/](https://github.com/Kuadrant/docs.kuadrant.io/)
+to update the Docs pointers to the tag or branch of the component repository that contains the updated documentation.
+file, to point to the tag or branch of the component repository that contains the updated documentation.
```
guicassolato, line 166:
```suggestion
```

guicassolato (approved):
Nice job, @didierofrivia!

rfcs/0000-kuadrant-release-process.md
```diff
@@ -0,0 +1,211 @@
+# Kuadrant Release Process
+
+- Feature Name: `kuadrant-release-process`
+- Start Date: 2024-01-11
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
```
guicassolato, line 6: Please don't forget to update these 😉

rfcs/0000-kuadrant-release-process.md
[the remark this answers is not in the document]
guicassolato:
```
Kuadrant vX.Y.Z = Kuadrant Operator vX.Y.Z =  Authorino Operator vA.B.C + Limitador Operator vD.E.F + DNS Operator vG.H.I + MGC Controller vJ.K.L
```

pehala (approved):

guicassolato (approved):

PR #51, opened 2024-01-23
Update RFC 0005 (Remove ManagedZone)

mikenairn:
We will no longer support the ManagedZone API as part of MGC/Kuadrant. This API was added at a time when we expected to create and manage many hosted zones in the target DNS provider. This is no longer the case and has no valid position in the future plans for a unified kuadrant.

### DIFF
--- a/rfcs/0005-single-cluster-dnspolicy.md
+++ b/rfcs/0005-single-cluster-dnspolicy.md
@@ -36,8 +36,6 @@ metadata:
 spec:
   providerRef:
     name: my-route53-credentials
-    namespace: my-gateways
-    kind: Secret
   targetRef:
     name: prod-web
     group: gateway.networking.k8s.io
@@ -79,7 +77,6 @@ metadata:
 spec:
   providerRef:
     name: my-route53-credentials
-    namespace: my-gateways
   endpoints:
     - dnsName: myapp.mn.hcpapps.net
       recordTTL: 60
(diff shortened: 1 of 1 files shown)


rfcs/0005-single-cluster-dnspolicy.md
```diff
@@ -196,16 +196,15 @@ DNSRecord:
...
+- ManagedZone API wil be removed and no longer supported as part of MGC/Kuadrant.
 
 ### DNSPolicy.spec.providerRef
 
-The `providerRef` field is mandatory and contains a reference to a resource that shows how DNSRecords will be reconciled.
-    - `spec.providerRef.name` - name of the provider resource
-    - `spec.providerRef.kind` - kind of resource, can be anything but only `Secret` or `Managedzone` will be reconciled by the `DNSPolicy` controller.
+The `providerRef` field is mandatory and contains a reference to a secret containing provider credentials.
+    - `spec.providerRef.name` - name of the provider resource.
+    - `spec.providerRef.kind` - kind of resource, must be `Secret`.
```
philbrookes, line 205: Any reason to keep this field then?

mikenairn: Yeah, i did think about that. We probably can just remove the kind and expect a secret name since we are saying now you must always set this and it must be a secret.

philbrookes: yep, sounds good to me!

mikenairn: removed

pehala (approved):

philbrookes (approved):
/lgtm

PR #55, opened 2024-02-09
Distributed DNS (RFC 0008)

philbrookes:

PR #70, opened 2024-03-04
RFC 00010 multi gateway shared DNS

maleck13:
Replaces https://github.com/Kuadrant/architecture/pull/55


rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,263 @@
...
+
+Enable the DNS Operator to manage DNS for one or more hostnames across one or more clusters.  Remove the requirement for a central OCM based "hub" cluster or multi-cluster control plane in order to enable multi-cluster DNS.
+
+# Motivation
+[motivation]: #motivation
+
+The DNS operator has limited functionality for a single cluster, and none for multiple clusters without leveraging OCM. Having the DNS operator capable of supporting the full feature set of DNSPolicy in these situations significantly eases on-boarding users to leveraging `DNSPolicy` across both single and multi-cluster deployments and gives a single layer to work against rather than defining resources in hubs and then understanding how to distribute them across spoke clusters.
+
+# Diagrams
+[diagrams]: #diagrams
```
mikenairn, line 24: Are there supposed to be diagrams here?

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,263 @@
...
+## General Logic Overview
+
+### Kuadrant Operator
+
+The general flow in the Kuadrant operator follows a single path, where it will always output a valid DNSRecord, which specifies everything needed for the listener host of a given gateway on the cluster, and is unaware of the concept of distributed DNS. The core logic for doing this already exists.
+
+
+### DNS Operator / controller
+
+We will update the DNS Operator to leverage the external DNS provider logic (IE effectively use it as a library) and layer our multi-cluster changes ontop of this logic. Our hope is to keep what we are doing as compatible with external DNS as possible so that we can contribute / propose changes to external-dsn
```
mikenairn, line 86:
```suggestion
We will update the DNS Operator to leverage the external DNS provider logic (IE effectively use it as a library) and layer our multi-cluster changes on top of this logic. Our hope is to keep what we are doing as compatible with external DNS as possible so that we can contribute / propose changes to external-dns
```

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,263 @@
...
...the relevant records for that DNS name from the provider and store them in the DNSRecord status. Next it will update the affected endpoints in that record set and validate the record set has no "dead ends" before writing the changes back to the DNS Provider's zone. Once the remote write is successful, it will then re-queue the DNSRecord for validation. The validation will be the same validation as previously done before the write. It will pull the zone and ensure that the endpoints from its own local DSNRecord are reflected correctly. If the validation is successful, the controller will mark this in the status of the DNSRecord and will re-queue the validation for ~15 (example) minutes later (a stable state verification). If validation is unsuccessful, it will re-queue the DNSRecord for validation rapidly (5 seconds for example). With each re-queue after an unsuccessful validation, it will add a random amount of "jitter" time to increase the chance it moves out of sync with any other actors. Each time it re-queues, it will mark this in the status (see below). At any point, if there is a change to the on-cluster DNSRecord, this back off and validation will be reset and started again.
```
mikenairn, line 87: What's the thinking behind adding the list of zone endpoints to the status? Is that just for visibility i.e. this list should eventually be the same on all DNSRecord instances giving you a full view of the world on any cluster?

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,263 @@
...
...writes some part of a zone/record with a stale view. This can happen if two or more controllers attempt to update the remote zone at the same time as each controller may have read the zone before another clusters has executed it's write allowing the zone to be updated without the knowledge of other actors that have also done a read. During this type of timing clash, effectively the last controller to write will have its changes stored. When it does validation, its changes will still be present and it will revert to a stable cycle. The other controllers involved, will see their changes missing in the remote zone during their validation check and so will pull and update the zone again (setting a new validation with a random jitter applied to increase the chances they don't clash again). Again the last controller to write will see its changes. So with this we can predict a worst case scenario of (num clashing controllers * (validation_loop + jitter)). However with adding in the jitter, it is likely that this will be a shorter period of time as multiple clusters will fall out of sync and should resolve their state within the min-max requeue interval rather than only ever one at a time.
```
mikenairn, line 89: Might be worth noting here that some DNS providers, depending on how they work, also try and mitigate against this kind of scenarios. Google for instance requires that during any update you first delete the existing record and you must do this by providing the exact contents of that record. In this case if it was out of date it would error and fall fall out of the reconcile loop and try again.

We still however need to do what is described here for cases where the provider does not do this.

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+## Orphan Mitigation
+
+It is possible for an orphan set of records to be created if for example a cluster is removed and the controller on that cluster is not given time to clean up. This is not something we currently mitigate against directly in the controller. However, you can define health checks tht would remove these orphans from the DNS response. Additionally, there is future work that leverages a "heart beat" that should allow each alive controller to see and remove a dead controllers records within a given criteria.
```
mikenairn, line 60:
```suggestion
It is possible for an orphan set of records to be created if for example a cluster is removed and the controller on that cluster is not given time to clean up. This is not something we currently mitigate against directly in the controller. However, you can define health checks that would remove these orphans from the DNS response. Additionally, there is future work that leverages a "heart beat" that should allow each alive controller to see and remove a dead controllers records within a given criteria.
```

rfcs/0008-distributed-dns.md
maleck13: Yes I think it is useful to be able to see what is the state that the controller is working from and allow comparision with controllers in other clusters

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,263 @@
...
+
+We can take the UID of that namespace and apply a hashing algorithm to result in 6-7 character code which can be used as the clusterID for the local cluster.
+
+#### Ensure local records are accurate
+
+When the local DNSRecord is updated, the DNS Operator will ensure those values are present in the zone by interacting directly with the DNS Provider. As mentioned above it will pull the zone, merge in its changes and write back the result before requeuing the resource for validation. With each requeuing it will add a random amount of time to attempt to fall out sync with any conflicting controllers.
+
+#### Cleanup
+
+When a DNS Policy is marked for deletion the kuadrant operator will delete all relevant kuadrantRecords.
```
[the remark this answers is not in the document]
maleck13: yes thanks that was left over

rfcs/0008-distributed-dns.md
maleck13: being updated

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,270 @@
...
+
+
+## Orphan Mitigation
+
+It is possible for an orphan set of records to be created if for example a cluster is removed and the controller on that cluster is not given time to clean up. This is not something we currently mitigate against directly in the controller. However, you can define health checks that would remove these orphans from the DNS response. Additionally, there is future work that leverages a "heart beat" that should allow each alive controller to see and remove a dead controllers records within a given criteria.
```
ficap, line 64: `However, you can define health checks that would remove these orphans from the DNS response.` -> In the end this may lead to serving of orphaned ip instead of NXDOMAIN if all of the other DNSPolicies from other clusters are removed (or how does the dns provider handle situations when there are no healthy endpoints left?).

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,270 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+
+## General Logic Overview
+
+### Kuadrant Operator
+
+The general flow in the Kuadrant operator follows a single path, where it will always output a valid DNSRecord, which specifies everything needed for the listener host of a given gateway on the cluster, and is unaware of the concept of distributed DNS. The core logic for doing this already exists.
```
ficap, line 85: what if DNSRecord object created from DNSPolicy is manipulated directly? Does kuadrant-operator reconcile it back to the desired state? if not and "multicluster is active" I think the dns-operator can easily be confused as the record structure may not be compliant with either strategy.

rfcs/0008-distributed-dns.md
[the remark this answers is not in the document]
maleck13: right this assumes a converged state is possible. It will work if there is a common state they are all trying to reach. Please see conflict detection below for how we will handle situations where it is clear we are not reaching that state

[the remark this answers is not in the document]
maleck13: No not all records in zone just any shared records (IE multiple values)

rfcs/0008-distributed-dns.md
maleck13: it returns all IPs if no healthy IPs left

rfcs/0008-distributed-dns.md
maleck13: Yes this will be part of reconciliation. IE watch the DNSRecord and bring back to the right state

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,270 @@
...
+
+The suggestion [here](https://groups.google.com/g/kubernetes-sig-architecture/c/mVGobfD4TpY/m/nkdbkX1iBwAJ) is to use the UID of the `kube-system` namespace, as this [cannot be deleted easily](https://groups.google.com/g/kubernetes-sig-architecture/c/mVGobfD4TpY/m/lR4SnaqpAAAJ) - so is practically speaking indelible.
+
+We can take the UID of that namespace and apply a hashing algorithm to result in 6-7 character code which can be used as the clusterID for the local cluster.
+
+#### Ensure local records are accurate
+
+When the local DNSRecord is updated, the DNS Operator will ensure those values are present in the zone by interacting directly with the DNS Provider. As mentioned above it will pull the zone, merge in its changes and write back the result before requeuing the resource for validation. With each requeuing it will add a random amount of time to attempt to fall out sync with any conflicting controllers.
```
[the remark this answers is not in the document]
maleck13: There is always a requeue. It is either a short requeue or a long requeue.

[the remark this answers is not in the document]
maleck13: I have also updated this now to call out using external-dns plan

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,274 @@
...
+### Configuration Updates
+There will be a new configuration option that can be applied as runtime argument (to allow us emergency configuration rescue in case of an unexpected issue) to the kuadrant-operator:
+- `clusterID`  (see below for how this is used)
+  - When absent, this is generated in code (more information below). A unique string to identify this cluster apart from all other clusters.
+
+### Changes to DNS Policy reconciler (kuadrant operator)
+
+This will primarily need to be made aware that it will need to propagate the DNS Policy health check block into any DNS Records that it creates or updates using the provider specific fields.
+
+It will also need to update the format of the records in the DNSRecord it creates. We will drop the use of a guid from the DNS name and instead simple use a prefix (klb) for example. This will mean multiple gateways with the same or different names can now update the record set if their gateway shares a listener with a common host name.
```
mikenairn, line 149: Also needs to add the `clusterID` as an annotation/label to any DNSRecords it creates. Should we add that here?

mikenairn: Starting to wonder if `ownerID` shouldn't just be added as an optional field to the  DNSRecord.Spec since it, along with "rootHost", are going to be important when it comes to multicluster.

mikenairn:
Couple of questions/thoughts, but nothing major.

/lgtm

rfcs/0008-distributed-dns-assets/dns-record.jpg
ficap, line 1: app.a.b.com -> lb.hash.app.a.b.com -> NXDOMAIN
AFAIK should be app.a.b.com -> lb.app.a.b.com

rfcs/0008-distributed-dns.md
ficap: ok, then if the orphan is the latest record it gets served (instead of NXDOMAIN) despite health checks, right?

rfcs/0008-distributed-dns.md
maleck13: might avoid specifics of how it is passed for now

rfcs/0008-distributed-dns.md
ficap: another fact that arises from not-mitigating orphans is that the strategy can't be changed for this hostname if orphan is not manually removed. (actually this may be a feature)

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,274 @@
...
...ds for that DNS name from the provider and store them in the DNSRecord status. Next via the external-dns plan, it will update the affected endpoints in that record set and validate the record set has no "dead ends" before writing the changes back to the DNS Provider's zone. Once the remote write is successful, it will then re-queue the DNSRecord for validation. The validation will be the same validation as previously done before the write. It will build a plan and if the plan based on the remote zone. If the plan is empty of changes the validation is successful. The controller will then mark this in the status of the DNSRecord and will re-queue the validation for ~15 (example) minutes later (a stable state verification). If validation is unsuccessful, it will re-queue the DNSRecord for validation rapidly (5 seconds for example). With each re-queue after an unsuccessful validation, it will add a random amount of "jitter" time to increase the chance it moves out of sync with any other actors. Each time it re-queues, it will mark this in the status (see below). At any point, if there is a change to the on-cluster DNSRecord, this back off and validation will be reset and started again.
```
ficap, line 91: `It will build a plan and if the plan based on the remote zone.` what??

maleck13: hmm obviously got distracted when writing that :) Will update to make more sense

maleck13:
@ficap I think your change requests are complete now

rfcs/0008-distributed-dns.md
```diff
@@ -0,0 +1,274 @@
...
+The controller will need to work through the DNS Record and ensure that from hostname to A (or external CNAME) records is structurally sound and that any dead ends are removed, to do this, the DNS Operator will need to first read what is published in the provider zone.
+
+#### Removing the finalizer from DNSRecord
+
+When a DNSRecord is reconciled, the DNS Operator should apply a finalizer to it.
+
+When a DNSRecord is being removed, the following must be successfully completed before the finalizer can be removed:
+- Remove the local clusters records and targets from the relevant zone
+- Perform a prune on the relevant root host
+- Apply the results of the prune to the DNS Provider
```
ficap, line 192: also delete healthchecks if needed (how to detect it - no records left?)

maleck13: This is a good point. @philbrookes any thoughts on when we delete a health check? I would think it is part of the dead end check? IE if there is no endpoints for a given rootHost then remove?

david-martin:
@philbrookes I noticed rfc 8 is already on main while looking through a couple other prs https://github.com/Kuadrant/architecture/blob/main/rfcs/0008-kuadrant-release-process.md

rfcs/0008-distributed-dns.md
maleck13: had to change the rfc number which is why this shows as outdated

ficap:
@maleck13 I think it would be also beneficial to list here all cases that can happen during plan creation and how they should be resolved. So that we can discuss possible ambiguous cases prior to implementation here.

maleck13:
> @maleck13 I think it would be also beneficial to list here all cases that can happen during plan creation and how they should be resolved. So that we can discuss possible ambiguous cases prior to implementation here.

Can you give an example of what you are expecting

ficap:
> > @maleck13 I think it would be also beneficial to list here all cases that can happen during plan creation and how they should be resolved. So that we can discuss possible ambiguous cases prior to implementation here.
> 
> Can you give an example of what you are expecting

Something like the "use cases to ensure" in this issue https://github.com/Kuadrant/dns-operator/issues/49 . Basically have a set of rules that would apply during plan creation.

rfcs/0009-distributed-dns.md
```diff
@@ -0,0 +1,291 @@
+# Distributed DNS Load Balancing
```
guicassolato, line 1: '0009' [already taken](https://docs.kuadrant.io/architecture/rfcs/0009-defaults-and-overrides/).

PR #85, opened 2024-03-13
Add info on release of example alerts and dashboars

david-martin:
Part of https://github.com/Kuadrant/kuadrant-operator/issues/448

Folders and initial example dashboards added in https://github.com/Kuadrant/kuadrant-operator/pull/461

No process changes needed to allow for this as the source bundle is already included in kuadrant-operator releases.


didierofrivia (approved):
🙏🏼

PR #89, opened 2024-05-16
include istio resources

maleck13:
minor update to call out the Istio resouces we use to integrate with Istio provided gateways


guicassolato (approved):
LGTM

PR #91, opened 2024-06-11
RFC-0002: Remove filter_state from the explicitly supported attributes

alexsnaps:
I've kept the anchor unchanged

### DIFF
--- a/rfcs/0002-well-known-attributes.md
+++ b/rfcs/0002-well-known-attributes.md
@@ -76,7 +76,7 @@ These two subsets of attributes are unified into a single set of well-known attr
 The next sections specify the well-known attributes organized in the following groups:
 - [Request attributes](#request-attributes)
 - [Connection attributes](#connection-attributes)
-- [Metadata and filter state attributes](#metadata-and-filter-state-attributes)
+- [Metadata attribute](#metadata-and-filter-state-attributes)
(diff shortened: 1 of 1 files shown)


rfcs/0002-well-known-attributes.md
```diff
@@ -76,7 +76,7 @@ These two subsets of attributes are unified into a single set of well-known attr
 The next sections specify the well-known attributes organized in the following groups:
 - [Request attributes](#request-attributes)
 - [Connection attributes](#connection-attributes)
-- [Metadata and filter state attributes](#metadata-and-filter-state-attributes)
+- [Metadata attribute](#metadata-and-filter-state-attributes)
```
guicassolato, line 79:
```suggestion
- [Metadata attribute](#metadata-attribute)
```

guicassolato: Although arguably the metadata attribute (singular) gives you a collection of other (not so well-known) attributes and values (plural.) So I wouldn't mind calling this section **Metadata attributes** if you want.

alexsnaps: >I've kept the anchor unchanged

As mentioned above, I had left the anchor unchanged, so that anything referencing that part wouldn't result in a broken link... So what do we do?

PR #93, opened 2024-06-27
rfc for removal of managed zone API

### LINKED ISSUE #67: RFC: Remove Managed Zone from DNSPolicy
https://github.com/Kuadrant/multicluster-gateway-controller/issues/715

Follows up on https://github.com/Kuadrant/architecture/pull/30

maleck13:
@jsmolar @mikenairn @philbrookes 

fixes #67


rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,97 @@
...
+
+If these fields are not present the status of the DNSPolicy will be updated to reflect the issue (more on status below). Only if these fields are present and listeners that match the `DOMAIN_NAME` are present will DNSRecords be created. If no record is created for a given secret, this will be considered a valid configuration (listeners may be added after a secret has been created and we shouldn't assume an error in this case).
+
+
+## Secret label selector
+
+The reason for using a label selector rather than a direct reference are as follows:
+
+- It will allow a single DNSPolicy to be applied to multiple dns providers (specific configuration being now present in the secret)
+- It will allow for multiple root domains (IE *.b.com and *.a.com can exist on a single gateway) to be used and controlled by DNSPolicy (currently not possible with managed zone)
```
david-martin, line 50: What's the impact on the controllers in terms of complexity, having to support these 1:M relationships to providers and domains?

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,97 @@
...
+DOMAIN_NAME: *.a.b.com
+
+```
+
+If these fields are not present the status of the DNSPolicy will be updated to reflect the issue (more on status below). Only if these fields are present and listeners that match the `DOMAIN_NAME` are present will DNSRecords be created. If no record is created for a given secret, this will be considered a valid configuration (listeners may be added after a secret has been created and we shouldn't assume an error in this case).
+
+
+## Secret label selector
+
+The reason for using a label selector rather than a direct reference are as follows:
```
pehala, line 47: I am missing reasoning why label selector are better than allowing multiple credential reference directly (i.e. credentials being a list of references to secrets).

Personally, I think this particular use case is better suited for the direct specification in the policy, than to have dynamic secret selectors (it is easier to work with as a user, easier to validate, everything can be seen from the spec alone)

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,97 @@
...
+
+ZONE_ID: somezone
+DOMAIN_NAME: *.a.b.com
+
+```
+
+If these fields are not present the status of the DNSPolicy will be updated to reflect the issue (more on status below). Only if these fields are present and listeners that match the `DOMAIN_NAME` are present will DNSRecords be created. If no record is created for a given secret, this will be considered a valid configuration (listeners may be added after a secret has been created and we shouldn't assume an error in this case).
+
+
+## Secret label selector
```
pehala, line 45: I think this is description of a implementation we want to use instead of a description of a feature or capability we want. I would rather see description of a feature first and then proposals of how to implement it

pehala (commented):
My biggest problem with this RFC is that it is bundling together two things that are not neccessarily connected in my eyes:
1) Removal of the ManagedZoned and replacing the content of credential secret
2) Adding ability to have multiple credentials for DNSPolicy

rfcs/0011-removal-managed-zone.md
maleck13: This is the ref level? 
```
This is the technical portion of the RFC. Explain the design in sufficient detail ...
```

rfcs/0011-removal-managed-zone.md
maleck13: I think this is worth thinking about. We have also been debating that. Label selectors are very common method for allow mutliple objects to be referenced, however a list type would also be an option.

maleck13:
> My biggest problem with this RFC is that it is bundling together two things that are not neccessarily connected in my eyes:
> 
> 1. Removal of the ManagedZoned and replacing the content of credential secret
> 2. Adding ability to have multiple credentials for DNSPolicy

Sure I can see that. But we have to remove the managedzone ref as part of this work and replace it with a secretRef anyway, so it seemed a reasonable point to question if we should go with a single secret or allow multiple to remove a limitation the DNSPolicy has currently.

rfcs/0011-removal-managed-zone.md
maleck13: There should be very little impact as it is one credential to one domain in one provider. So effectively rather than load a single managedzone and then load a credential and reconcile the dns record, it would instead be a case of load the credential(s) and reconcile the associated dns records. So a change in the DNSPolicy would trigger the reconcile of each associated DNSRecord

rfcs/0011-removal-managed-zone.md
pehala: How about writing it as an alternative solution in the RFC, properly exploring and investigating pros and cons of each solution

pehala:
> Sure I can see that. But we have to remove the managedzone ref as part of this work and replace it with a secretRef anyway, so it seemed a reasonable to question if we should go with a single secret or allow multiple as part of this work allowing us to remove a limitation the DNSPolicy has currently and avoid two breaking changes one after the other. So this was really an attempt to be somewhat efficient and remove the need for another RFC.

Sure, my problem with this is that while removing managedzone is simple enough and probably will go through without much debate. Multiple credentials is a new feature and might require more discussion. When I looked through the PRs I didnt pay much attention to this RFC because I thought it handles just the removal.

I would either split this into two RFC or rescope this RFC to focus mainly on the new feature and as part of that we will remove ManagedZone

maleck13:
> > Sure I can see that. But we have to remove the managedzone ref as part of this work and replace it with a secretRef anyway, so it seemed a reasonable to question if we should go with a single secret or allow multiple as part of this work allowing us to remove a limitation the DNSPolicy has currently and avoid two breaking changes one after the other. So this was really an attempt to be somewhat efficient and remove the need for another RFC.
> 
> Sure, my problem with this is that while removing managedzone is simple enough and probably will go through without much debate. Multiple credentials is a new feature and might require more discussion. When I looked through the PRs I didnt pay much attention to this RFC because I thought it handles just the removal.
> 
> I would either split this into two RFC or rescope this RFC to focus mainly on the new feature and as part of that we will remove ManagedZone

Yeah I think that is reasonable. And to be fair the concept emerged as part of thinking about removal. So I guess we can say the original RFC was intended to just be removal of the MZ but our thinking evolved and so re-focusing the RFC to be around supporting multiple providers and domains from a single policy is a reasonable approach or removing that part into a separate RFC 👍

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,97 @@
...
+The reason for using a label selector rather than a direct reference are as follows:
+
+- It will allow a single DNSPolicy to be applied to multiple dns providers (specific configuration being now present in the secret)
+- It will allow for multiple root domains (IE *.b.com and *.a.com can exist on a single gateway) to be used and controlled by DNSPolicy (currently not possible with managed zone)
+
+Each secret discovered will result in a separate DNSRecord for any targeted listener in the gateway that matches or is a subdomain of the DOMAIN_NAME field specified in the secret using a longest prefix match strategy. Each DNSRecord will have a direct reference to the secret that resulted in its creation. 
+
+## Secret deletion / modification
+
+When a secret is deleted there will be no action taken by the DNS Controller. When a secret is updated the controller will trigger a reconcile of the policy.
```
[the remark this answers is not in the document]
maleck13: This info should def make it into the status, I agree. So triggering a reconcile and allowing the controller to error out and report in the status quickly due to a missing secret makes sense 👍

rfcs/0011-removal-managed-zone.md
maleck13: This has been removed now this RFC is just about removal of managed zone with the support for multiple provider record mirroring as something separate

rfcs/0011-removal-managed-zone.md
maleck13: This has been updated now and uses and array

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,97 @@
...
+
+The key user impact here is to avoid users having to create additional none policy `Kuadrant APIs` in order to be able to leverage the `DNSPolicy` API. 
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+You will no longer need or be able to create a `ManagedZone` resource. It will be expected that the zone already exists in your provider of choice. Now you will only create a credential via a Kubernetes secret. This secret will be where any additional configuration is also applied such as the zone id to target and the root domain in that zone to use. Removing the managed zone API will reduce API calls to the DNS provider and reduce events and load on the server and kuberentes API server.
+You will no longer be limited to a single root domain or single provider per policy. The managed zone was a single reference, however we want to move to leverage a label selector which will mean multiple credentials can be used for a single policy.
```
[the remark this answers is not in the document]
maleck13: separated out now

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,107 @@
...
+  providerRefs
+    - name: myawsroute53
+
+```
+Initially this array will have a limit of 1. The reason for an array is we expect to want to allow multiple providers secrets to be specified. This allows us to control the number of secrets and allow for possible expansion in the future.
+
+Where we use the existing credential we will now load a list of credentials and iterate through each of them and create a specific DNSRecord instance for the target DNS provider populating that DNSRecord with a reference to its required credential. The DNS Operator will pick up on these records in the same way it already does except it will load the secret directly rather than via the ManagedZone. The provider type is already specified as part of our existing secret definition:
+
+```
+type: kuadrant.io/aws
```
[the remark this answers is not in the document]
maleck13: it will not be picked up

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,107 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+change the DNSPolicy to have local object reference array for provider secrets in the same namespace.
+
+```
+spec:
+  providerRefs
+    - name: myawsroute53
```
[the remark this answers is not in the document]
maleck13: This is mentioned below. The DNS Operator will report an error in the existing sychronised status condition and this will be mirrored back into the DNSPolicy existing `recordConditions`

maleck13:
@mikenairn could I get a review on this

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,107 @@
...
+
+```
+
+ZONE_ID: somezone
+DOMAIN_NAME: *.a.b.com
+
+```
+
+Validation of the secret structure will happen in the dns operator rather than doing it once in the kuadrant operator and again in the dns operator. If the secret doesn't exist or if the required fields are not present the status of the DNSRecord will be updated to reflect the issue and the DNSPOlicy will reflect this status condition in its existing `recordConditions` section as it does now.
+DNSPolicy controller will only validate that the specified DOMAIN exists in the target gateway. If it does not exist it will mark this in the status of DNSPolicy.
```
mikenairn, line 61: What does this mean? Is the policy controller having to read data fro the providerRef(secret)?

rfcs/0011-removal-managed-zone.md
```diff
@@ -0,0 +1,107 @@
...
+Any issues updating/contacting a target DNS provider will be reflected in the existing `recordConditions` status via the existing `synchronised` condition.
+
+## Secret structure
+
+under the data field we will expect two new fields to be present
+
+```
+
+ZONE_ID: somezone
+DOMAIN_NAME: *.a.b.com
```
mikenairn, line 56: The only reason i see needing the DOMAIN_NAME in the secret would be if we intend to look it up, much like we do currently for [ManagedZones](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dns_helper.go#L42), is that the intention? Are we still only creating a DNSRecord if we have a `providerRef` that has a matching domain set? If that is the case, domain name here would not have the wildcard on it, it should match the zone domain name `a.b.com`.

maleck13: Yes I forgot to remove this after our conversation online... Will update. I think we just create the DNSRecord and allow the DNS Opertor to do the work and pass back the result.

philbrookes (approved):

Issue #48, opened 2024-01-19
Distributed tracing on data plane


alexsnaps:
As request come into the cluster, it'd be nice to enable distributed traces from the Gateway onward, throughout the different Kuadrant components and eventually to the user's application fronted by Kuadrant.

 - [x] Kuadrant/kuadrant-operator#399
 - [x] Kuadrant/authorino-operator#167
 - [x] Kuadrant/wasm-shim#47
 - [x] Kuadrant/wasm-shim#48
 - [x] Kuadrant/limitador#86
 - [x] Kuadrant/limitador#275
 - [x] Kuadrant/limitador-operator#130
 - ~Add some global way (in the `Kuadrant` CR?) to configure the underlying components for them to push spans to the OTLP collector.~ Moved to https://github.com/Kuadrant/kuadrant-operator/issues/731
 - ~Have us (within wasm, e.g. pre auth RL?) start the trace and is there a way to have envoy propagate that traceparent to ext`-auth?~ Moved to https://github.com/Kuadrant/wasm-shim/issues/59
 - ~Look into adding "well-known" fields into our span to "trace" them back to the user's domain: RLP, AP, …~ Moved to https://github.com/Kuadrant/architecture/pull/96

david-martin:
@alexsnaps @adam-cattermole for the sake of gh issue hygiene, and assuming at this time there are no issues in this epic progressing, how about closing this out with the tasks that were completed & marking this as Done?
The remaining tasks could be moved to a new epic and prioritised accordingly.

adam-cattermole:
Sure, I think that makes sense - breaking out the pieces / decisions around how we expose tracing at the kuadrant level

david-martin:
Closing as done with follow up work captured in:
- https://github.com/Kuadrant/kuadrant-operator/issues/731
- https://github.com/Kuadrant/wasm-shim/issues/59
- https://github.com/Kuadrant/architecture/pull/96

PR #96, opened 2024-07-09
RFC: Tracing Sampling Strategy Rules

david-martin:
This RFC is an attempt to capture the thoughts and considerations around the bullet item "Look into adding "well-known" fields into our span to "trace" them back to the user's domain: RLP, AP, …" from https://github.com/Kuadrant/architecture/issues/48

TODO: 
- [ ] Get answers to unresolved questions & update the doc accordingly
- [ ] Investigate 'prior art'

### DIFF
--- a/rfcs/0000-tracing-sampling-strategy-rules.md
+++ b/rfcs/0000-tracing-sampling-strategy-rules.md
@@ -1,0 +1,89 @@
+# Tracing Sampling Strategy Rules
+
+- Feature Name: tracing_sampling_strategy_rules
+- Start Date: 2024-07-09
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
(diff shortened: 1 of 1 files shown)


Boomatang:

alexsnaps:
Wrt to the `prior art`, you might want to look into [head- vs. tail-based sampling](https://newrelic.com/blog/how-to-relic/distributed-tracing-and-sampling#head-based-sampling-overview) in tracing. But I'm somewhat confused about two main use cases: 

 - Error (somewhere late in the game something "fails", which that would need definition already - an exception/panic/err would be already complicated bu what about e.g. "too slow" request) and as a user I'd certainly would want to have that trace available
 - Enabling tracing on a "late in the game known attribute", e.g. trace for user `bob`, how would limitador know that pre-auth ratelimiting phase would need to send traces for ... `bob` that we'd only know about in the auth phase?

I _think_ this tries to resolve tail-based sampling... If so, do we want to do that? Now that being said, there is still much more we can do to be a better citizen in that space.

PR #97, opened 2024-07-23
RFC: Observability Policy

R-Lawton:


rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+          limitador: 
+            logMode: development  
+    
+    tracing:
+      component:  
+        limitador:
+          endpoint: rpc://tempo.tempo.svc.cluster.local:4317
+          insecure: true
+          tags: tag1, tag2
+          strategyRules:
```
R-Lawton, line 61: incorporating potential other RFC https://github.com/Kuadrant/architecture/pull/96

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+      kuadrant:   
+        enableService: true
+        port: 8084 
+  logging:
+    component:
+      authorino: 
+        logLevel: debug      
+```
+### Status
+The status of the Observability CR will not be the observability stack is in a "healthy" state i.e Prometheus and Grafana is up and running. It should be the status of only the things that we contribute for example is new Logging and Tracing now in place or is the console plugin responding. We will not be taking responsibility for aspects we don't have control over.
```
david-martin, line 156:
```suggestion
The status of the Observability CR will not be the observability stack is in a "healthy" state i.e Prometheus and Grafana is up and running. It should be the status of only the things that we contribute for example is new Logging and Tracing now in place. We will not be taking responsibility for aspects we don't have control over. 
```
console plugin removed from spec

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+In terms of how the information supplied in the Observability CR will get passed to the other Kuadrant components, multiple options have been brought up:
+
+directly - like setting flags or configuration directly on the deployments of the different Kuadrant components.
+indirectly -  Passing the information to Authorino & limitador via the Authorino & Limitador CRs
+indirectly -  Passing the information to Authorino & limitador in the form of there own Observability CRs
+
+The best approach would be the indirect approach, meaning once the Observability CR is updated the information is passed to the relevant component CR. For example the tracing section in the Authorino CR spec would be updated with the required endpoints and other configuration in the Observability CR, this would then be updated in the Authorino CR.
```
david-martin, line 168: Agreed.
The direct option could break easily when authorino or limitador change something in the deployment for example.
This does mean the available options that can be included in the Observability spec are limited to what's exposed via the Authorino & Limitador CRs (but changes can be requested)

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+* Single source of truth: Rather then having multiple crs to check what the current configuration is theres a single source of truth. Preventing users from accidentally changing values by mistake
+
+## Other options:
+An other option that has been investigated which is very similar to above, is having observability configuration as a element of the kuadrant CR spec. The majority of the work itself would be largely the same with operators having to move configuration to the Kuadrant CR and having new observability features use the kuadrant CR as the source of truth. 
+
+The reason why we should go with the above method is Abstraction. The Kuadrant CR quite quickly can get "muddied" with observability and be harder to read and maintain also losing its main purpose; being the install kick of and maintainer for the kuadrant operators as well as the single source of truth aspect. From a user point of view having a users have to change configuration in 3+ separate crs in some cases is tedious and not slow.
```
david-martin, line 204:
```suggestion
The reason why we should go with the above method is Abstraction. The Kuadrant CR quite quickly can get "muddied" with observability and be harder to read and maintain also losing its main purpose; being the install kick of and maintainer for the kuadrant operators as well as the single source of truth aspect. From a user point of view having a users have to change configuration in 3+ separate crs in some cases is tedious and slow.
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+          deep: true
+        dns-operator:
+          enableService: true
+          port: 8080   
+
+    alerts:
+      namespace: my-amazing-namespace
+      component:
+        authorino:
+          operator-level: true
```
david-martin, line 104: lower camelCase field names would fit better with k8s standards

```suggestion
          operatorLevel: true
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+| Tracing               | Authorino operator, Limitador operator                    | **component** , **endpoint** , **tags**  , **insecure**, **strageyRules**  |
+| Metrics               | Kuadrant operator, Authorino, Limitador, DNS Operator     | **component** , **enableService** , **port** , **deep**                                            |
+| Alerts *              | Kuadrant operator, Authorino, Limitador                   | **namespace** , **component**, **enable**                                                                 |
+| Dashboards *          | Kuadrant operator, Authorino, Limitador                   | **namespace** , **component**, **enable**                                                                 |
+| Other 3rd Party *     | e.g Kiali                                                 | **enable** bool                                                                                                               |
+
+###### **Note**: Observability pieces with a * denotes these are post v1 milestone
+
+### Example CR with everything
```
david-martin, line 37: I wonder if it would clarify expectations better if you say this example is an ideal example if every component had every option available, and was configurable via the same type of API?
Reiterating about the evolution of the spec via versioning could re-inforce that.
Perhaps a set of guidelines the spec could follow is enough at this stage (which is kind of what is shown anyways)

rfcs/0000-observability-api.md
david-martin: Maybe a note to say this section will depend on work in that RFC?

david-martin (changes requested):

rfcs/0000-observability-api.md
R-Lawton: Sure ill add that in

rfcs/0000-observability-api.md
R-Lawton: Good catch i'll remove it

rfcs/0000-observability-api.md
R-Lawton: Yes that a good point to acknowledge

rfcs/0000-observability-api.md
R-Lawton: Sure i can do that

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,223 @@
...
+
+```yaml 
+
+apiVersion: kuadrant.io/v1alpha1
+kind: observability
+spec:
+      logging:
+        component:
+          authorino:
+            logLevel: debug
```
maleck13, line 47: We should probably have the same property as standard rather than `logLevel` and `logMode`

R-Lawton: I can see now its not as obvious in the example. Loglevel and logmode are two separate options let me update that to make that more obvious

R-Lawton: Also see copy paste error to

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,234 @@
...
+
+
+# Unresolved questions
+
+1. Do we want a value where a person just wants everything "on", they want all the observing. Where a value a the root of the observability section can be set to say give me all you got?
+2. Do we go with a observability policy instead of the proposed configuration style CR like whats being suggested? TLDR of what that would look like is quite similar to the observability CR in the fact that the observability piece is pulled out but instead of there just being one source of truth with the current method, and having the fine grained capabilities be in the one location. Instead you would have for most use cases a single observability policy that would be attached at the Kuadrant component level, in there you would have a Kuadrant level view i.e if i enable tracing here its enabled for all the components that can handle it. From there if a component needs some special configuration a separate policy can be created and attached to that specific component instead.
```
R-Lawton, line 187: @Boomatang Does that cover for the most part what we were discussing this morning?

R-Lawton: Im now leaning to this solution being the one to go with but want to get other peoples thoughts before i write a full rfc on it

Boomatang: I think that covers what we talked about.

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,235 @@
+# RFC Template
```
guicassolato, line 1:
```suggestion
# Observability API
```

maleck13:
So discussed with @R-Lawton in person, but there seems to be some redundancy in the configuration and needs a user to understand the names and configuration options for underlying components.

I think we can simplify this down somewhat into a two phase approach

1) offer log and tracing options (but they will only affect Authorino and Limitador) in phase 1 and be as simple as:

```
logging:
   level: debug
tracing:
  common tracing config applied to both limitador and authorino
metrics:
  enabled: false (removes service monitors) enabled by default
```
Where by default it is set to `error` level for logging and a default logging config plus , metrics defaulted to on. 

We can then expand beyond this in a second phase 

add comming configmap for expressing observability values that can be loaded as env vars (logging for example) so that it changes all components including the operators . Add in new options around alerts and dashboards as needed.

The status of the kuadrant resource can reflect what has been configured:

```
logging:
   authorino: debug
   limitador: debug
tracing:
  ... 

```

WDYT @Boomatang @R-Lawton

R-Lawton:
> 1. offer log and tracing options in the kuadrant CR (but they will only affect Authorino and Limitador) in phase 1 and be as simple as:
> 
> ```
> logging:
>    level: debug
> tracing:
>   common tracing config applied to both limitador and authorino
> metrics:
>   enabled: false (removes service monitors) enabled by default
> ```
> 
> Where by default it is set to `error` level for logging and a default logging config plus , metrics defaulted to on.

@maleck13 

By default the log level is info, what are your thoughts about changing it to error?

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,148 @@
...
+        endpoint: rpc://tempo.tempo.svc.cluster.local:4317
+        tags: tag1, tag2
+    metrics: 
+        enable: false
+    logging:
+      logLevel: debug      
+```
+
+### Status
+The status of the Kuadrant CR will not be the observability stack is in a "healthy" state i.e Prometheus and Grafana is up and running. It should be the status of only the things that we have control over. The status will show what components the user will have configured for example logging setup for Authorino and Limitador. We will not be taking responsibility for aspects we don't have control over.
```
KevFan, line 62:
```suggestion
The status of the Kuadrant CR will not be whether the observability stack is in a "healthy" state, i.e., Prometheus and Grafana is up and running. It should be the status of only the things that we have control over. The status will show what components the user have configured, for example, the logging setup for Authorino and Limitador. We will not be taking responsibility for aspects we don't have control over. 
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+In terms of how the information supplied in the Kuadrant CR will get passed to the other Kuadrant components, multiple options have been brought up:
+
+- directly - like setting flags or configuration directly on the deployments of the different Kuadrant components.
+- indirectly -  Passing the information to Authorino & Limitador via the Authorino & Limitador CRs
+- indirectly -  Passing the information to Authorino & Limitador in the form of there own Observability CRs
```
KevFan, line 85:
```suggestion
- indirectly -  Passing the information to Authorino & Limitador in the form of their own Observability CRs
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,148 @@
...
+* Single source of truth: Rather then having multiple CR's to check what the current configuration is theres a single source of truth. Preventing users from accidentally changing values by mistake
+
+## Other options:
+An other option that has been investigated which is very similar to above, is having observability configuration as its own API. The majority of the work itself would be largely the same with operators having to move configuration to the Kuadrant CR and having new observability features use the Kuadrant CR as the source of truth. 
+
+The reason why we should not go with the above method is reliance. The Observability CR will rely on the Kuadrant CR being present and we moving away from that model. From a user point of view having a users have to change configuration in 3+ separate crs in some cases, is tedious and slow.
+
+If we don't decide to any of these options the user will have to manually in multiple places add there configuration of their desired observability stack which can result in poor user experience, mistakes being made and values not being tracked properly.
```
KevFan, line 134:
```suggestion
If we don't decide to any of these options, the user will have to manually add the configuration of their desired observability stack in multiple places, which can result in poor user experience, mistakes being made and values not being tracked properly.
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,148 @@
...
+
+# Summary
+[summary]: #summary
+
+This is a proposal for adding the ability to configure aspects of observability for Kuadrant as a whole to the already existing Kuadrant custom resource.
+
+# Motivation
+[motivation]: #motivation
+
+Users of Kuadrant (Platform engineers and or Site reliability engineers) will want a way to be able to configure different aspects of observability for Kuadrant without having to configure things at the component level themselves. Providing users this ability allows them to not have to have a deep knowledge of the inner workings of the components as well as integrate Kuadrant observability pieces with their own observability stack or Openshift user workload monitoring. Resulting in a fully comprehensive observability view all configured in one location.
```
KevFan, line 16:
```suggestion
Users of Kuadrant (Platform engineers and or Site reliability engineers) will want a way to be able to configure different aspects of observability for Kuadrant without having to configure things at the component level themselves. Providing users this ability allows them to integrate Kuadrant observability pieces with their own observability stack or Openshift user workload monitoring without needing a deep knowledge of the inner workings of the components, resulting in a fully comprehensive observability view all configured in one location.
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,148 @@
...
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+In terms of how the information supplied in the Kuadrant CR will get passed to the other Kuadrant components, multiple options have been brought up:
+
+- directly - like setting flags or configuration directly on the deployments of the different Kuadrant components.
+- indirectly -  Passing the information to Authorino & Limitador via the Authorino & Limitador CRs
+- indirectly -  Passing the information to Authorino & Limitador in the form of there own Observability CRs
+
+The best approach would be the indirect approach, meaning once the Kuadrant CR is updated, the information is passed to the relevant component CR. For example the tracing section in the Authorino CR spec would be updated with the required endpoints and other configuration from the Kuadrant CR. This does mean that the spec exposed is whats in the Kuadrant component CR's but new changes can be requested and implemented in said components.
```
KevFan, line 87:
```suggestion
1. directly - like setting flags or configuration directly on the deployments of the different Kuadrant components.
2. indirectly -  Passing the information to Authorino & Limitador via the Authorino & Limitador CRs
3.  indirectly -  Passing the information to Authorino & Limitador in the form of there own Observability CRs

The best approach would be the approach 2), an indirect approach, meaning once the Kuadrant CR is updated, the information is passed to the relevant component CR. For example, the tracing section in the Authorino CR spec would be updated with the required endpoints and other configuration from the Kuadrant CR. This does mean that the spec exposed is whats in the Kuadrant component CR's, but new changes can be requested and implemented in said components.
```

Maybe use a numbered list instead since the description seems to describe the first indirect approach specificially? 🤔

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,149 @@
...
+
+**As of the time of writing this, no the original suggestion solves the issues we need right now.**
+
+#### Work thats needed
+The indirect approach allows for not much if any changes to the Authorino operator and the Limitador operator etc. The bulk of the work that would be needed would be in the Kuadrant operator.
+
+The changes will come into full affect through a phased approach due to the nature of some aspects not being available yet. The phased approach can follow the versioning syntax that k8s like v1beta1 or v1alpha1 and be portrayed in the CRD.
```
KevFan, line 116:
```suggestion
The indirect approach allows for few, if any, changes to the Authorino operator and the Limitador operator etc. The bulk of the work that would be needed would be in the Kuadrant operator.

The changes will come into full effect through a phased approach due to the nature of some aspects not being available yet. The phased approach can follow the versioning syntax that k8s like v1beta1 or v1alpha1 and be portrayed in the CRD. 
```

rfcs/0000-observability-api.md
```diff
@@ -0,0 +1,149 @@
...
+[prior-art]: #prior-art
+
+Manually adding configuration to Kuadrant operator crs.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Currently we only have configuration for Logging, Tracing and Metrics. Post v1 the plan is to add alerts, dashboards and potentially other 3rd party like Kiali.
+
+* For the logging piece this is currently only for Authorino and Limitador and not there operators or the Kuadrant operator as these require flags to be changed so in phase two we can look at what work needs to be done to make this cohesive. For example being able to change a config and that gets percolated down to all the components .
```
KevFan, line 148:
```suggestion
* For the logging piece, this is currently only for Authorino and Limitador and not their operators or the Kuadrant operator as these require flags to be changed so in phase two we can look at what work needs to be done to make this cohesive. For example being able to change a config and that gets percolated down to all the components .
```

KevFan (commented):
Looks generally good, I've left some minor suggestions 👍

david-martin (approved):

KevFan (approved):
Looks good to me 👍

Boomatang:

maleck13:
Wanted to call out that there is a dicussion and investigation outlined here about some of these areas 
https://github.com/Kuadrant/kuadrant-operator/issues/1130
We have multiple "lower level" things that a user may want to configure. We want to have a consistent experience around how these are achieved. So let hold this PR until that is completed

david-martin:
Since the Kuadrant CRD is v1, I don't see any avenue to trying things out with new APIs exposed through the Kuadrant CR.
For example, introducing some alpha API, seeing how it gets used and adapting it based on feedback.

@maleck13 @Boomatang @KevFan @R-Lawton @guicassolato What are your thoughts on exploring new APIs through a separate CRD. In this case, that would be an Observability CRD at v1alpha1.
There are a few ways that API could go then.
It could eventually evolve into a v1 (as the Observability CRD).
It could eventually fizzle out and be removed.
Or it could be merged into the Kuadrant CR.

I worry that the outcome of discussions and linked topics in https://github.com/Kuadrant/kuadrant-operator/issues/1130 could end up stifling or delaying potential improvements and new features while we agonise over getting the perfect API.

maleck13:
The Kuadrant CRD is not v1 at the moment (correctly or incorrectly) https://github.com/Kuadrant/kuadrant-operator/blob/main/config/install/configure/standard/kuadrant.yaml

david-martin:
Apologies, you're right.
Well that does change my argument above.
I think there's still an argument for using separate CRDs so as not to complicate versioning of the Kuadrant CRD.
There may be a conversation about how the Kuadrant CRD is versioned in general, and if/when it makes it to v1.

I'd still be interested in peoples thoughts on the above argument about using a separate v1alpha1 CRD to introduce this new feature.

david-martin:
> Why yet another CRD

I like CRDs. To me, they make configuration feel more controlled and kube native.
A big part of that is having a spec, field validation and taking advantage of built in kubectl cli args like output formatting and templating.

That being said, if there's push back on a CRD approach, the configmap solution can certainly work as you've outlined, and in a safe way.

> If you are wanting to get this in front of user fast

There's no users screaming out for this.
It's something I think can make installation easier overall for some users.
For now, I'm content to let the discussion continue here and in https://github.com/Kuadrant/kuadrant-operator/issues/1130, before refining this RFC.

KevFan:
@david-martin Personally, I'm leaning towards just prototyping with the Kuadrant CR regardless of its current API version since the RFC "consenus" (at the moment) is to use the Kuadrant CR to configure Observability. I don't think we should care too much on complicating the versioning of the `Kuadrant` CR given this 🤔 Say even if it was at `v1`, we can add a `v2alpha1`, which makes a new alpha version that allows us to experiment with it 🤔 If the consenus had been to use separate CRD from the start, then that would make complete sense. 

I like the general conventions set by the k8s api for introducing any changes to their api and the understanding of "the internal representation of an API object is decoupled from any one API version":
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

We can prototype this as a non-breaking change to either `v1beta1` as an optional field since the spec is currently empty (until there's an new upstream release) or better yet introduce a `v1beta2` that we prototype with but keep `v1beta1` as the stored version and the only version we generally recommend. If we're saying users can try this feature, use `v1beta2` but knowing that it's a feature that is in the early stages and can have breaking changes until it's the recommended version 🤔 Achieving this is probably not "easy", but may be the "simpler/cleaner" way eventually. Eventually we'll reach a `v1betaN` version which will be promoted as `v1` of Kuadrant CR if that's the goal.

I'm guessing what may add to this complexity and what would affect deciding on whether we want to or not use the Kuadrant CR to prototype with this approach is possibly the upgrades and release differences with upstream and downstream 🤔 (e.g. upstream with many releases now at `v1betaN`, and removed `v1beta1`/no longer reconcile/convert from this version, but downstream takes in the latest release at `v1betaN`)

If the goal is to just experiment quick on observability, then a separate CRD makes sense that I'm not totally against anyhow.

PR #109, opened 2024-11-25
UPDATE: change the status of RFC 0006 to Withdrawn

Boomatang:
The RFC was reverted and there was no way to show this before.


Boomatang:
@alexsnaps As this RFC was reverted after being excepted I feel there needs to be a way to show that. I have added a status field with `Withdrawn` as the status. What do you think.

alexsnaps:
Sounds good with me... maybe we should update that [section](https://github.com/Kuadrant/architecture?tab=readme-ov-file#the-rfc-lifecycle) if not this time, if it ever happens again?

alexsnaps (approved):

PR #110, opened 2024-11-27
RFC for mTLS

david-martin:
Follow on from investigation in https://github.com/Kuadrant/kuadrant-operator/issues/1049 and a possible automated way to enable and manage mTLS for users instead of the manual documented way in https://github.com/Kuadrant/kuadrant-operator/pull/1054

### DIFF
--- a/rfcs/0012-mtls.md
+++ b/rfcs/0012-mtls.md
@@ -1,6 +1,7 @@
 # RFC Template
 
 - Feature Name: `mtls`
+- Status: **Implemented**
 - Start Date: 2024-11-27
 - RFC PR: [Kuadrant/architecture#110](https://github.com/Kuadrant/architecture/pull/110)
 - Issue tracking: [Kuadrant/kuadrant-operator#1049](https://github.com/Kuadrant/kuadrant-operator/issues/1049)
@@ -9,6 +10,7 @@
 [summary]: #summary
 
 MTLS support for communication between the Wasm plugin in the gateway pod, and Authorino & Limitador.
+Currently scoped to Istio-based gateway providers only.
 
 # Motivation
 [motivation]: #motivation
@@ -20,14 +22,16 @@ This aligns with a 'zero trust' architecture.
 [guide-level-explanation]: #guide-level-explanation
 
 Update the Kuadrant CR to have a new configuration option `mtls` for users to configure if mtls should be enabled for either authorino or limitador.
-The reason for having separate flags for either component is to allow flexibility where the resource overhead of tls may be considered too high when enabled for a specific component e.g. for checking limits on each request in limitador.
+A global `enable` flag controls the feature, with optional per-component overrides to allow flexibility where the resource overhead of tls may be considered too high when enabled for a specific component e.g. for checking limits on each request in limitador.
+When `enable` is `true`, both components default to enabled unless explicitly overridden with `false`. When `enable` is `false`, mTLS is disabled regardless of per-component values.
 The new config in the Kuadrant CR would look like this:
 
 ```yaml
-apiVersion: Kuadrant.io/v1alpha1
-kind: kuadrant
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
 spec:
   mtls:
+    enable: true
     authorino: true
     limitador: true
 ```
@@ -35,6 +39,8 @@ spec:
 When enabled for a specific component, the following changes are performed by the kuadrant-operator and/or component operators. This may require trickling down configuration to other operator resources, like the Authorino or Limitador resources:
 
 ## Enabling the Istio sidecar proxy in that component
+
+Note: This feature currently requires Istio as the gateway provider.
 
 This can be done by adding the `sidecar.istio.io/inject=true` label to the component deployment template.
 The pod spec is updated dynamically by Istio to inject the `istio-proxy` sidecar into the pod.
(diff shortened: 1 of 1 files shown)


rfcs/0012-mtls.md
```diff
@@ -0,0 +1,106 @@
...
+kind: kuadrant
+spec:
+  mtls:
+    authorino: true
+    limitador: true
+```
+
+When enabled for a specific component, the following changes are performed by the kuadrant-operator and/or component operators. This may require trickling down configuration to other operator resources, like the Authorino or Limitador resources:
+
+## Enabling the Istio sidecar proxy in that component
```
guicassolato, line 37: We should probably add a section for Envoy Gateway as well.

rfcs/0012-mtls.md
```diff
@@ -0,0 +1,106 @@
...
+The reason for having separate flags for either component is to allow flexibility where the resource overhead of tls may be considered too high when enabled for a specific component e.g. for checking limits on each request in limitador.
+The new config in the Kuadrant CR would look like this:
+
+```yaml
+apiVersion: Kuadrant.io/v1alpha1
+kind: kuadrant
+spec:
+  mtls:
+    authorino: true
+    limitador: true
```
guicassolato, line 32: Although this could work for a first implementation, I wonder if it's clear to the user that the certs will be provisioned via [Secret discovery service (SDS)](https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret).

I can also see how non-mesh users could want this but by configuring certs by other means though. If we don't support such other means right away, should the API be more explicit about using SDS perhaps?

david-martin: So the available options right now for mTLS seem quite limited (from reading this recent article https://blog.howardjohn.info/posts/mtls-kubernetes/)
There's either a sidecar approach (which this uses), ambient mode, a CNI approach (which doesn't exist yet), or DIY.
Is the abstraction too low if SDS is mentioned in the config?
What about an mTLS mode of 'mesh' being the initial option, with that meaning we'll integrate with the mesh if it's available i.e. if you're using Istio.

rfcs/0012-mtls.md
david-martin: I don't know what to add for that.
The Istio approach takes advantage of Istio sidecar capabilities to enable mTLS in the kuadrant pods.
If Istio isn't there, would that be some other solution outside of Envoy Gatway completely?

rfcs/0012-mtls.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
```
maleck13, line 34: I think it would be good to call out that if we don't support mTLS configuration for a gateway provider, it will be reflected in the status of the kuadrant resource. Also given how important this can be for users, kuadrant should probably not become ready until the specified mTLS can be handled or the mTLS section is removed.

maleck13 (commented):
Some minor improvements suggest but, I like this as an initial pass at allowing configuration of mTLS when using Istio. Avoids users having to mess with resources owned by the kaudrant operator and also kuadrant users not needing to understand the innards of Envoy. It also seems relatively easy to extend with things like a specific CA etc.

rfcs/0012-mtls.md
Boomatang, line 32: I think there is a design issue here. More so when we think of the whole CR and just this feature. As this feature only has two components that is affected this will be harder to visualize, but I will try explain it from a generic features point of view.

```yaml
spec:
  feature:
    activate: false # optional, boolean, defaults to false
    enabled: [] # optional, list of strings
    disabled: [] # optional, list of strings
```
Under the feature all fields are optional to ensure backwards compatibility. The feature will not be activated on less the `activate` field is set to true. This allows the feature to be enabled or disabled without changing the configuration for that feature. A feature would have a list of defaults components that would be affected by the feature being enabled. In this case that would be two, `authorino` & `limitador`, but this would not always be the case. 

If the user wanted to disable the feature on a component lets say limitador they would add the following spec.
```yaml
spec:
  feature:
    activate: true 
    disabled: ['limitador']
```
In this example only limitador will not have the feature enabled, but also if any new components are added to the defaults list that will be auto enabled. Being auto enabled by default may not be what the user wants. That is where the `enabled` list comes in. If the user wants to only have the feature enabled for authorino no matter what components are add in the future they could use the following spec.
```yaml
spec:
  feature:
    activate: true 
    enable: ['authorino']
```
The enable list would be an explicit list. Only components listed would be affected. This ensures that behaviour would not change if a component is added to the support list later in the future. However, if there are a lot of components that is supported by default filling this list in could be tedious, but as it is a string list there is nothing stopping use from supporting keywords like `'defaults'`. An example of this being used could be service monitors whereby default we only add them to the operands. A user many also add them to the operator that they care about. The below example shows that looks like.
```yaml
spec:
  feature:
    activate: true
    enable: 
      - 'defaults'
      - 'authorino-operator'
```

So advantage that I see with doing this design are.
- All features have the same basic interface.
- New components don't require CRD updates to be supported. The change will only be required with the operator.
- The common interface make documentation simpler

david-martin: I like this approach.
Are the enabled & disabled lists mutually exclusive?

maleck13: One thing I am thinking here is we could go with an opinionated mTLS but ensure we allow users to still change values directly if needed. So when reconciling, we would assert that the section is there but not overwrite any new config for example. 
So out of the box you get a working version that uses SDS if you want more than that you can configure it directly and we wont overwrite it

Boomatang:

rfcs/0012-mtls.md
```diff
@@ -0,0 +1,106 @@
...
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Manual configuraton is an option (investigated in https://github.com/Kuadrant/kuadrant-operator/issues/1049 and documented in https://github.com/Kuadrant/kuadrant-operator/pull/1054)
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- What components will create and manage the resources and configuration detailed above?
+- What configuration options makes sense to bubble up via the Limitador and Authorino components, if any?
```
laurafitzgerald, line 106: Discussion around this happening -> https://groups.google.com/a/redhat.com/g/kuadrant-dev/c/rbsY4kDaSnE
And at the community call -> https://docs.google.com/document/d/1Dkcqu1MgCJlvJHF_iCYibPNsnSHu1aIeG5-WoqlKtGU/edit?tab=t.0#heading=h.5a78l4pgyj90
Will relay the outcome here via update with an answer to this question.

maleck13:
@david-martin is this still relevant? Be good for it reflect the current state and be merged

PR #111, opened 2024-12-11
Update architecture reference to align with v1

david-martin:
- Fixed links to policy reference pages
- updated references to how auth integration with provider is done to align with current approach (wasm, similar to rate limiting)
- removed references to OCM for multi cluster
- added link to multi cluster DNS RFC PR
- added Envoy Gateway as an alternative dependency to Istio
- updated various diagrams to remove references to Istio resources that are no longer used, and to reference Envoy Gateway as an alternative
- updated the multi cluster diagram to increase font size in various boxes

images are smaller resolution, but still should render fine in 'Rich diff' and on the docs site default width


david-martin:
@maleck13 when you get a chance, I'd like to get this landed on main and backported to the v1.0.x branch of the docs site

maleck13 (approved):

PR #112, opened 2025-01-13
RFC: Standardize the Kuadrant Spec

Boomatang:
Start RFC that outlines an Standardized API design for Kuadrant Spec

There is a number of RFC attempting to add new features to Kuadrant that would be configured through the Kuadrant CR. There has being attempts to add feature before the V1 release, and these features got removed due to the messy API design that was used.

Before any new feature work that would be configured in the kuadrant CR, we should agree on a common API design as this will affect how features are designed to work.

### DIFF
--- a/rfcs/0000-standardize_the_kuadrant_spec.md
+++ b/rfcs/0000-standardize_the_kuadrant_spec.md
@@ -1,0 +1,238 @@
+# Standardize the Kuadrant Spec
+
+- Feature Name: Standardize the Kuadrant Spec
+- Start Date: 10/01/2025
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC is trying to describe a common API design that can be used when adding feature to Kuadrant that are configurable thought the Kuadrant CR.
+
+# Motivation
+[motivation]: #motivation
+
+As we are move forward with the development of Kuadrant, we are wanting to add new features to the product.
+Some of these features will be configurable by the user, the Kuadrant CR is an expected place for this to happen. 
+Before starting to add these configurable features, a standardize way of defining features in the Kuadrant CR should be agreed upon. 
+This RFC lays the groundwork for an API design which the new features will follow.
+By having a standardized API design, future development can move faster as the API design will have the groundwork already done.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The core of the design is made up of four elements. 
+`spec.feature_a` is the name of the feature that is being configured in that section.
+`spec.feature_a.configure` is a boolean field which tells the kuadrant operator to use the configuration defined in the Kuadrant CR for the feature. 
+This is done to allow us the flexibility to have features active by default.
+`spec.feature_a.enable` is a list of strings that represent the components the configuration will be applied too.
+Having it a list of strings allows us to extend or reduce the number of accepted values without needing to modify the CRD. 
+With the use of strings as values means keywords can also be defined, for example `"default"` could be shorthand for a sub set of components affected by the feature.
+`spec.feature_a.disable` is a list of strings that repenting the components that the configuration will *not* be applied too.
+Again shorthand keywords can be used here.
+```yaml
+spec:
+  feature_a:
+    configure: true # optional, boolean, defaults to false
+    enable: [] # optional, list of strings
+    disable: [] # optional, list of strings
+```
+This may not be the prefect design, and in the following use case of possible future kuadrant feature will show how far the design and concept can be pushed.
+
+## mTLS
+
+This feature has an [RFC](https://github.com/Kuadrant/architecture/pull/110/files) which describes how the feature works.
+The example here shows how that configuration would work in the standard being outlined in this RFC.
+At time of writing there is a discussion on wheather the feature should be active by default. 
+In this example it is assumed the feature will be active by default.
+
+There are only two components that are outlined in the related RFC as being affected by the feature, `authorino` and `limitador`.
+This means that by default the feature would be enabled for both elements.
+If the user only wants the feature enable for one of the two there is two approaches the user can take.
+The first is to be explicit about which component they want the feature enabled for.
+```yaml
+spec:
+  mtls:
+    configure: true
+    enable: ["authorino"]
+```
+Or they could be explicit about which components they do **not** want the feature enabled for.
+```yaml
+spec:
+  mtls:
+    configure: true
+    disable: ["limitador"]
+```
+Both of the above configurations achieve the same effect.
+If in the future this feature applies to a third component, the behaviour would not be the same.
+In the first example the new component would not get the configured, while in the second example it would be configured.
+
+If the user was wanting to fully disable the feature, as this example assumes active by default, they would use the following configuration.
+```yaml
+spec:
+  mtls:
+    configure: true
+    disable: ["authorino", "limitador"]
+```
+Again if in the future if a third component was added, in the above configuration it may not be disabled by default.
+This can be addressed by using keywords, and `"defaults"` in this case.
+```yaml
+spec:
+  mtls:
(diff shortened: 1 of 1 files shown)


rfcs/0000-standardize_the_kuadrant_spec.md
```diff
@@ -0,0 +1,238 @@
...
+
+# Summary
+[summary]: #summary
+
+This RFC is trying to describe a common API design that can be used when adding feature to Kuadrant that are configurable thought the Kuadrant CR.
+
+# Motivation
+[motivation]: #motivation
+
+As we are move forward with the development of Kuadrant, we are wanting to add new features to the product.
```
maleck13, line 16:
```suggestion
As we move forward with the development of Kuadrant, we are wanting to add new features to the product.
```

rfcs/0000-standardize_the_kuadrant_spec.md
```diff
@@ -0,0 +1,238 @@
...
+
+There are only two components that are outlined in the related RFC as being affected by the feature, `authorino` and `limitador`.
+This means that by default the feature would be enabled for both elements.
+If the user only wants the feature enable for one of the two there is two approaches the user can take.
+The first is to be explicit about which component they want the feature enabled for.
+```yaml
+spec:
+  mtls:
+    configure: true
+    enable: ["authorino"]
```
maleck13, line 58: So from the start I was looking to see how this would be presented. I worry that this is overly verbose? If instead you had  a default for `mtls` of not enabled this would be the equivalent of leaving it blank or more sepecifically ```mtls:
       enable: []
```
so I am wondering is the configure part needed?

Comment (Boomatang) [reply rfcs/0000-standardize_the_kuadrant_spec.md:58]: Being verbose is on propose, it gives a clearer picture the user about what they are doing. In your example, if there is more then one feature in the kuadrant CR, lets `mtls` and `service monitors`. One feature is enabled by default and the second is disabled. What would a empty enable list mean on both configurations? 

The verbose nature shows the intend in a much clearer. It is not just about how easy it is to configure today, but also how easy would it be for a new system administrator to onboard after kuadrant is used by our users for years to come.  

From a technical point, I think an empty list would be dropped from the spec as I think we would need to do as omit if  empty so as not make breaking changes to API. If this is the case it would become harder to understand what is happening.

File: rfcs/0000-standardize_the_kuadrant_spec.md [hunk 3]
```diff
@@ -0,0 +1,238 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+The core of the design is made up of four elements. 
+`spec.feature_a` is the name of the feature that is being configured in that section.
+`spec.feature_a.configure` is a boolean field which tells the kuadrant operator to use the configuration defined in the Kuadrant CR for the feature. 
+This is done to allow us the flexibility to have features active by default.
+`spec.feature_a.enable` is a list of strings that represent the components the configuration will be applied too.
+Having it a list of strings allows us to extend or reduce the number of accepted values without needing to modify the CRD. 
+With the use of strings as values means keywords can also be defined, for example `"default"` could be shorthand for a sub set of components affected by the feature.
+`spec.feature_a.disable` is a list of strings that repenting the components that the configuration will *not* be applied too.
```
maleck13, line 32: I think this could be added at a later date if it was shown to be needed. We don't have that many components and so an enable list should not be too onerous on the user IMO

rfcs/0000-standardize_the_kuadrant_spec.md
```diff
@@ -0,0 +1,238 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+
+# Prior art
+[prior-art]: #prior-art
```
david-martin, line 211: I took inspiration for the grafana config from the kiali CR https://kiali.io/docs/configuration/kialis.kiali.io/
I've used the prometheus and alertmanager CRs a lot https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api-reference/api.md#monitoringcoreoscomv1
Both seem like reasonable prior art in this space.

rfcs/0000-standardize_the_kuadrant_spec.md
```diff
@@ -0,0 +1,238 @@
...
+Having it a list of strings allows us to extend or reduce the number of accepted values without needing to modify the CRD. 
+With the use of strings as values means keywords can also be defined, for example `"default"` could be shorthand for a sub set of components affected by the feature.
+`spec.feature_a.disable` is a list of strings that repenting the components that the configuration will *not* be applied too.
+Again shorthand keywords can be used here.
+```yaml
+spec:
+  feature_a:
+    configure: true # optional, boolean, defaults to false
+    enable: [] # optional, list of strings
+    disable: [] # optional, list of strings
```
david-martin, line 39: I think having both an 'enable' and 'disable' list would be confusing for the user.
Are they meant to be mutually exclusive?
Should only 1 or the other be set?
Or should I make sure all components are included across the union of both lists.

Boomatang: When you say mutually exclusive, which do you mean? 
1. you can only have the `enable` or `disable` lists.
2. a value can only be in `enable` or `disable`, but not in both.

rfcs/0000-standardize_the_kuadrant_spec.md
Boomatang: I can add them is as prior art. 

The kiali CR reference shows how out of control this can get. That is only for one product while we would need to be able to do that for many.

rfcs/0000-standardize_the_kuadrant_spec.md
Boomatang: You maybe right, but I would like to have a guide to what happens if the disable is required. I don't like the idea of updating the CRD with new fields in the future, if we can plan for it now.

rfcs/0000-standardize_the_kuadrant_spec.md
```diff
@@ -0,0 +1,238 @@
...
+  logging_level:
+    - level: "error"
+      configure: true
+      enable: ["kuadrant-operator"]
+      disable: []
+    - level: "debug"
+      configure: true
+      enable: ["kuadrant-operator"]
+      disable: []
+```
```
Boomatang, line 179: Thinking about this again. There might be some thing that we can do with CEL to ensure a component is only configured in one listed object.

rfcs/0000-standardize_the_kuadrant_spec.md
david-martin: Either.
Hence the confusion.
Also, what happens if i'm not exhaustive in what i put in both arrays?
For example:
```yaml
enable: ['authorino']
disable: ['limitador']
```
Do all **other** components have the feature enabled or disabled?

I'm not looking for an answer here (at least not yet).
Moreso highlighting that it's confusing to me.

rfcs/0000-standardize_the_kuadrant_spec.md
david-martin: > how out of control this can get

What makes you say it's out of control?
From the [developer docs](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-operator-resource), it looks like they have a good handle on managing configuration.
Perhaps it's different from those on the ground developing it, but from a user point of view, the docs have the entire spec [in one place](https://kiali.io/docs/configuration/kialis.kiali.io/), and detailed docs for the different sections, for example, https://kiali.io/docs/configuration/p8s-jaeger-grafana/grafana/

> That is only for one product while we would need to be able to do that for many.

Do you mean one component?
Kuadrant is one project, with multiple components in my view.

Boomatang:
Related conversation: https://kubernetes.slack.com/archives/C05J0D0V525/p1737380600067349

rfcs/0000-standardize_the_kuadrant_spec.md
Boomatang: I will do a bit of rewording around this. I am guess that the `# optional` is also causing some of this confusion.

PR #115, opened 2025-03-05
Open arch

alexsnaps:


rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+Think of a `MetapolicyController` being a pure function that takes the custom *Metapolicy* and the Gateway API network
+objects it targets as input, and outputs one or more Kuadrant Policies as a result.
+
+## Implementing a custom *Metapolicy*
+
+Fundamentally a *Metapolicy* isn't much different from a regular Kubernetes [Custom
+Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/), and as such will be
+managed by some controller. But unlike a regular [Kubernetes
+Controller](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#custom-controllers),
+it also needs to reconcile when changes to properties of the Gateway API it uses are observed. For that reason, the
```
thomasmaas, line 56:
```suggestion
it also needs to reconcile when it observes changes to properties of the Gateway API it uses. For that reason, the
```

Or is it bot the controller that observes the changes?

thomasmaas: In that case maybe: 

```suggestion
it also needs to reconcile when changes to properties of the Gateway API it uses are observed by the `KuadrantController`. For that reason, the
```

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+kind: Kuadrant
+metadata:
+  name: kuadrant-sample
+spec:
+  plugins:
+    userMetaPolicy:
+      deployment: embedded
+```
+
+The `UserMetaPolicy` will need to be packaged with a custom Docker image containing the plugin. It'll be automatically
```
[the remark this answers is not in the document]
alexsnaps: The `Plugin` in this case is the extension point that the Kuadrant controller needs to know about. It so happens that this extension is a "metacontroller" that effectively cares about changes to its own resources, e.g. `UserMetaPolicy` CRs _and_ the changes to relevant Gateway API resources (i.e. the DAG).

Now there is no reason the `Plugin` couldn't be more or something entirely different than a "metacontroller". It simply formalizes the API that such a plugin talks with the Kuadrant controller. That, and "sandboxes" it, so that plugins failing don't affect each other.

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+  rs.Labels["gateway"] = label
+  err = r.client.Update(ctx, rs)
+  if err != nil {
+    return reconcile.Result{}, fmt.Errorf("could not write MetaPolicy: %+v", err)
+  }
+
+  return reconcile.Result{}, nil
+}
+```
+It on differs in three ways from the original example:
```
maleck13, line 102:
```suggestion
It only differs in three ways from the original example:
```

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. The `Reconcile` signature takes an additional parameter, the `kuadrant *KuadrantContext`;
+1. It acts upon an hypothetical `user.MetaPolicy`;
+1. It uses the `evaluateExpression` to resolve the CEL expression `kuadrant.gatewaysFor(self)[0].metadata.name` instead
+   of using a fixed string for the label's value.
+
+Because that expression is evaluated upon creation, the `KuadrantRuntime` will make sure to reconcile whenever that cel
```
maleck13, line 109:
```suggestion
Because that expression is evaluated upon creation, the `KuadrantRuntime` will make sure to reconcile whenever that CEL
```

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+> Gateway API objects queried
+
+- [ ] Support existing `context`s in the CEL interpreter's `Env`
+- [ ] Find a way to express the `context`s present/set for each CEL
+
+
+### Deployment
+
+> [!IMPORTANT]
+> This would initially be deployed as their own processes running alongside the kuadrant controller. They'd communicate
```
maleck13, line 167: When you say along side. Does that mean in the same pod and as part of the same deployment object?

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+> though unix sockets, over gRPC & gRPC streams (for the eventing).
+>
+> Later, we could support different deployment models, as all that'd be needed is being able to use gRPC essentially.
+
+- [ ] Known location for plugins to live in; one per directory
+- [ ] Unix socket opened within the same location?
```
maleck13, line 173: Possibly use a volume mount of type emptyDir. Th https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

rfcs/0000-open-arch.md
maleck13: One consideration may be limiting the resources that such a container can have. I think we would also want to understand to protect the kuadrant operator from a bad plugin crashing and potentially causing the pod to get restarted

maleck13: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-restarts

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+    return reconcile.Result{}, fmt.Errorf("could not fetch MetaPolicy: %+v", err)
+  }
+
+  // Set the label if it is missing
+  if rs.Labels == nil {
+    rs.Labels = map[string]string{}
+  }
+	
+  // resolve the CEL expression using the `KuadrantContext`
+  label := kuadrant.evaluateExpression("kuadrant.gatewaysFor(self)[0].metadata.name")
```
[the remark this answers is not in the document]
alexsnaps: yes... where `gatewaysFor` would probably provide "overlays", as CEL calls them, to act upon certain types. I don't know actually that we can do it, in this case, for anything that has `targetRef`, a la structural typing... so, might have to be `gatewaysFor(self.targetRef)`, where that then resolves to a known-type for with there is an overlay.

rfcs/0000-open-arch.md
alexsnaps: I was, for now certainly, not looking into making this a "first class citizen" per se, so you'd probably docker compose the base image to add e.g. the `/plugins/myPlugin/myPlugin` executable to the kuadrant operator.

rfcs/0000-open-arch.md
```diff
@@ -0,0 +1,225 @@
...
+> - What is the impact of not doing this?
+
+# Prior art
+[prior-art]: #prior-art
+
+> [!todo]
+> Discuss prior art, both the good and the bad, in relation to this proposal.
+> A few examples of what this can include are:
+>
+> - Does another project have a similar feature?
```
david-martin, line 196: Similar to ext_proc in envoy proxy? https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter

PR #119, opened 2025-04-08
ADD: Resilient Deployment

### LINKED ISSUE #117: RFC Resilient Deployment
The end result for this work will add configuration to the kuadrant CR.

Boomatang:
closes: #117

### DIFF
--- a/rfcs/0000-resilient_deployment.md
+++ b/rfcs/0000-resilient_deployment.md
@@ -1,0 +1,182 @@
+# RFC Template
+
+- Feature Name: Resilient Deployment
+- Start Date: 2025-04-07
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#117](https://github.com/Kuadrant/architecture/issues/117)
+
+# Summary
+[summary]: #summary
+
+<!-- TODO: One paragraph explanation of the feature. -->
+
+# Motivation
+[motivation]: #motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected outcome? -->
+As our user move to deploying kuadrant into production environments there is a want to create more resilient deployments.
+This includes deploying multiply replicas of the data plane components, Authorino & Limitador.
+In the case of Limitador, this also includes being able to persist counters.
+
+As Kuadrant we want to provide a user experience that makes deploying a more resilient product possible.
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+<!-- Explain the proposal as if it was implemented and you were teaching it to Kuadrant user. That generally means: -->
+<!---->
+<!-- - Introducing new named concepts. -->
+<!-- - Explaining the feature largely in terms of examples. -->
+<!-- - Explaining how a user should *think* about the feature, and how it would impact the way they already use Kuadrant. It should explain the impact as concretely as possible. -->
+<!-- - If applicable, provide sample error messages, deprecation warnings, or migration guidance. -->
+<!-- - If applicable, describe the differences between teaching this to existing and new Kuadrant users. -->
+
+At the core, there are two different areas of focus required here.
+There is the resilience of the deployments, Authorino & Limitador, and there is the resilience of the counters used within Limitador.
+These two areas address the resiliency as whole.
+
+For the deployments there are a number of configurations required.
+- TopologyConstraints
+- PodDisruptionBudget
+- Resource Limits
+- Replicas
+With Authoring and Limitador there are different ways of configuring some of the features.
+
+The counters in Limitador can be persisted with external storage.
+- [Counter storages](https://github.com/Kuadrant/limitador/blob/main/doc/server/configuration.md#counter-storages)
+- [Limitador API](https://github.com/Kuadrant/limitador-operator/blob/main/api/v1alpha1/limitador_types.go#L203-L211)
+
+Configuration of the feature will be done in the Kuadrant CR.
+This is to make it simpler for the user.
+**API Design (WIP)**
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+spec:
+  resilience:
+    authorization: True
+    rateLimiting: True
+    counterStorage: {} # lifts storage struct from the limitador CR.
+```
+
+Equally important, if not more important will be relaying information back to the user on the status of the configuration.
+What the status block looks like is unknown currently.
+<!-- TODO: more understanding is required here -->
+
+## Behavior
+It is important that we understand what the behavior of this feature should be, and convey that behavior to the user clearly (documentation).
+There are a number of uses that needs to be covered in the first iteration of this feature.
+- New installations
+- Existing installations with zero custom configuration
+- Existing installations with custom configuration
+- Removal of configuration
+- SRE reaction events
+- The user wants more
+
+### New installations
+This is by far the simplest configuration.
+By default, the all resilience configuration will be blank.
+Thus, the feature will be disabled.
+One reason for this choice is area the counter storage requiring configuration.
+
+### Existing installations with zero custom configuration
+This act very much like a new installation. 
+The resilience configuration will be blank, thus, the feature is disabled.
+
+Once the feature is enabled the all the relevant resources will be created, with the kuadrant-operator having ownership where possible.
+
+### Existing installations with custom configuration
+In the case the installation has custom configuration nothing changes till the users defines the resilience feature in the Kuadrant CR.
+
+Once the user configures the resilience feature the kuadrant-operator will create any missing configurations, and take ownership of those configurations.
+The kuadrant-operator will not modify, extend, update, or take ownership of any existing configuration[^1].
+However, the status will reflect there is configuration that the kuadrant-operator doesn't owner, or there is configuration outside what the kuadrant-operator expects.
+
+[^1]: This true for any resource that is not already managed by the kuadrant-operator. 
+
+### Removal of configuration.
+In this case the user has configured the resilience features in the past, and is turning them off.
+Where the features are turned off, the kuadrant-operator will clean resources it owns.
+Even if the user has modified that configuration.
+
+This needs to be called out very clearly in the documentation.
+Why we do this is to simplify the reconciliation of configuration.
+
+### SRE reaction events
+From time to time SRE teams would need to change configurations for other to address issues.
+We should allow the user configure the deployment to a state that is we regard as not resilient.
+The one example that comes to mind is scaling replicas to zero for address some issue.
+The user should be allowed to that, but the Kuadrant CR should state the configuration is below expected spec.
+The below spec would be regarded as an error.
+
+### The user wants more
+The user knows more about their deployment than we ever can.
+We need to allow the user to modify the configuration to suit their needs.
+When a user modifies the configuration, the kuadrant-operator should not reconcile back the changes.
+
+If the user changes are below the minimum spec that we state, this should be reflected in the status.
+If the user changes the configuration at all it should be reflected in the status. 
+However, when the spec is more than what we regard as minimum the status should be a highlight not warning or error.
+How this is reflected in the status of the Kuadrant CR is unknown currently.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+<!-- This is the technical portion of the RFC. Explain the design in sufficient detail that: -->
+<!---->
+<!-- - Its interaction with other features is clear. -->
+<!-- - It is reasonably clear how the feature would be implemented. -->
+<!-- - How error would be reported to the users. -->
+<!-- - Corner cases are dissected by example. -->
+<!---->
+<!-- The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work. -->
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+<!-- Why should we *not* do this? -->
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+<!-- - Why is this design the best in the space of possible designs? -->
+<!-- - What other designs have been considered and what is the rationale for not choosing them? -->
+<!-- - What is the impact of not doing this? -->
+
+# Prior art
+[prior-art]: #prior-art
+
+<!-- Discuss prior art, both the good and the bad, in relation to this proposal. -->
+<!-- A few examples of what this can include are: -->
+<!---->
+<!-- - Does another project have a similar feature? -->
+<!-- - What can be learned from it? What's good? What's less optimal? -->
+<!-- - Papers: Are there any published papers or great posts that discuss this? If you have some relevant papers to refer to, this can serve as a more detailed theoretical background. -->
+<!---->
+<!-- This section is intended to encourage you as an author to think about the lessons from other tentatives - successful or not, provide readers of your RFC with a fuller picture. -->
+<!---->
+<!-- Note that while precedent set by other projects is some motivation, it does not on its own motivate an RFC. -->
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+<!-- - What parts of the design do you expect to resolve through the RFC process before this gets merged? -->
+<!-- - What parts of the design do you expect to resolve through the implementation of this feature before stabilization? -->
+<!-- - What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC? -->
+
+Should resilience be enabled by default? 
+While the simple answer would be yes, there is the issue of a configuration being required for counter persistence's.
+
+While the kuadrant-operator will not take ownership of existing configuration.
+What the sub operators would do is unknown at this stage if the feature is enabled via their CRs.
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+<!-- Think about what the natural extension and evolution of your proposal would be and how it would affect the platform and project as a whole. Try to use this section as a tool to further consider all possible interactions with the project and its components in your proposal. Also consider how this all fits into the roadmap for the project and of the relevant sub-team. -->
+<!---->
+<!-- This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related. -->
+<!---->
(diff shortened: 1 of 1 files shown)


rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected outcome? -->
+As our user move to deploying kuadrant into production environments there is a want to create more resilient deployments.
+This includes deploying multiply replicas of the data plane components, Authorino & Limitador.
+In the case of Limitador, this also includes being able to persist counters.
+
+As Kuadrant we want to provide a user experience that makes deploying a more resilient product possible.
```
maleck13, line 21: would we call it an opinionated data plane configuration? To a certain extent we are setting it up based on our own understanding of these components and I would say common topology setups (AZ for example).

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+<!-- - If applicable, describe the differences between teaching this to existing and new Kuadrant users. -->
+
+At the core, there are two different areas of focus required here.
+There is the resilience of the deployments, Authorino & Limitador, and there is the resilience of the counters used within Limitador.
+These two areas address the resiliency as whole.
+
+For the deployments there are a number of configurations required.
+- TopologyConstraints
+- PodDisruptionBudget
+- Resource Limits
```
maleck13, line 42: Perhaps we would limit oursleves to resource requests and leave the actual limits open

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+- Replicas
+With Authoring and Limitador there are different ways of configuring some of the features.
```
maleck13, line 44:
```suggestion
With Authorino and Limitador there are different ways of configuring some of the features.
```

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+
+The counters in Limitador can be persisted with external storage.
+- [Counter storages](https://github.com/Kuadrant/limitador/blob/main/doc/server/configuration.md#counter-storages)
+- [Limitador API](https://github.com/Kuadrant/limitador-operator/blob/main/api/v1alpha1/limitador_types.go#L203-L211)
+
+Configuration of the feature will be done in the Kuadrant CR.
+This is to make it simpler for the user.
+**API Design (WIP)**
```
maleck13, line 52: I think the design is a good start.

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+Equally important, if not more important will be relaying information back to the user on the status of the configuration.
+What the status block looks like is unknown currently.
+<!-- TODO: more understanding is required here -->
+
+## Behavior
+It is important that we understand what the behavior of this feature should be, and convey that behavior to the user clearly (documentation).
+There are a number of uses that needs to be covered in the first iteration of this feature.
+- New installations
```
maleck13, line 70: Are the first two the same? Reconcile a set on known resources in a known state into the correct state? I guess the difference is in initial creation of the resources but the reconciles should be the same

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+- Existing installations with zero custom configuration
+- Existing installations with custom configuration
+- Removal of configuration
+- SRE reaction events
+- The user wants more
+
+### New installations
+This is by far the simplest configuration.
+By default, the all resilience configuration will be blank.
```
maleck13, line 79: It might not be blank for a new configuration. If I install the operator and then create the kuadrant CR for first time with the resilience block filled in?

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+Thus, the feature will be disabled.
+One reason for this choice is area the counter storage requiring configuration.
+
+### Existing installations with zero custom configuration
+This act very much like a new installation. 
+The resilience configuration will be blank, thus, the feature is disabled.
+
+Once the feature is enabled the all the relevant resources will be created, with the kuadrant-operator having ownership where possible.
```
maleck13, line 87:
```suggestion
Once the feature is enabled all the relevant resources will be created, with the kuadrant-operator having ownership where possible.
```

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Existing installations with custom configuration
+In the case the installation has custom configuration nothing changes till the users defines the resilience feature in the Kuadrant CR.
+
+Once the user configures the resilience feature the kuadrant-operator will create any missing configurations, and take ownership of those configurations.
+The kuadrant-operator will not modify, extend, update, or take ownership of any existing configuration[^1].
+However, the status will reflect there is configuration that the kuadrant-operator doesn't owner, or there is configuration outside what the kuadrant-operator expects.
```
maleck13, line 94:
```suggestion
However, the status will reflect there is configuration that the kuadrant-operator doesn't own, or there is configuration outside what the kuadrant-operator expects.
```

maleck13: How will we know there is additonal configuration objects such as PodDisruptionBudgets? or are you only referring to configuration applied within objects we already own?

rfcs/0000-resilient_deployment.md
maleck13: I still think this is a simple configuration as we are going from nothing to something well defined. Is this perhaps what you were trying to convey. No kuadrant CR means no existing configuration ?

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+- Existing installations with custom configuration
+- Removal of configuration
+- SRE reaction events
+- The user wants more
+
+### New installations
+This is by far the simplest configuration.
+By default, the all resilience configuration will be blank.
+Thus, the feature will be disabled.
+One reason for this choice is area the counter storage requiring configuration.
```
maleck13, line 81: not sure I follow this sentence

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+### Existing installations with zero custom configuration
+This act very much like a new installation. 
+The resilience configuration will be blank, thus, the feature is disabled.
+
+Once the feature is enabled the all the relevant resources will be created, with the kuadrant-operator having ownership where possible.
+
+### Existing installations with custom configuration
+In the case the installation has custom configuration nothing changes till the users defines the resilience feature in the Kuadrant CR.
+
+Once the user configures the resilience feature the kuadrant-operator will create any missing configurations, and take ownership of those configurations.
```
maleck13, line 92:
```suggestion
Once the user configures the resilience feature, the kuadrant-operator will create any missing configuration, and take ownership of them.
```

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+The kuadrant-operator will not modify, extend, update, or take ownership of any existing configuration[^1].
+However, the status will reflect there is configuration that the kuadrant-operator doesn't owner, or there is configuration outside what the kuadrant-operator expects.
+
+[^1]: This true for any resource that is not already managed by the kuadrant-operator. 
+
+### Removal of configuration.
+In this case the user has configured the resilience features in the past, and is turning them off.
+Where the features are turned off, the kuadrant-operator will clean resources it owns.
```
maleck13, line 100: I think this is reasonable. What is the intent of the user when removing the values in the CR? After all they added them in the first place, then modified to add additional configuration. 

Adding the initial configuration signals the desire to have the components configured as described. Adjusting the configuration in a supported way, signals a need or requirement of the environment. Removing the configuration signals that I want you to remove the what was set up. 
If we express any explicit values in this, for example we allow setting CPU requests (just an example) then that should always be reconciled to that value. In our current design we are not being "explicit" about what the configuration should be, instead we are setting up what we consider to be a base line for a resilient deployment  then expecting the user to use that as a jumping off point.

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+Even if the user has modified that configuration.
+
+This needs to be called out very clearly in the documentation.
+Why we do this is to simplify the reconciliation of configuration.
+
+### SRE reaction events
+From time to time SRE teams would need to change configurations for other to address issues.
```
maleck13, line 107:
```suggestion
From time to time SRE teams may need to change some configuration to address specific issues.
```

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+We should allow the user configure the deployment to a state that is we regard as not resilient.
+The one example that comes to mind is scaling replicas to zero for address some issue.
```
maleck13, line 109:
```suggestion
The one example that comes to mind is scaling replicas to zero to address some issue.
```

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+The user should be allowed to that, but the Kuadrant CR should state the configuration is below expected spec.
+The below spec would be regarded as an error.
+
+### The user wants more
+The user knows more about their deployment than we ever can.
+We need to allow the user to modify the configuration to suit their needs.
+When a user modifies the configuration, the kuadrant-operator should not reconcile back the changes.
```
maleck13, line 116: This is true to a point. If they delete the PDB, I would expect we put it back for example. If we have an explicit value in the kuadrant CRD that should also be reconciled back IMO. but for this feature in general I agree. But I wanted to call out the deletion as I don't think we want a situation where a PDB is deleted by accident and in order to get it back I first have to remove the config from the Kuadrant CR

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+
+If the user changes are below the minimum spec that we state, this should be reflected in the status.
+If the user changes the configuration at all it should be reflected in the status.
```
maleck13, line 119: I wonder can we just generalise this in some way.  I think given the user is making a change we don't have to say "its below minimum" I think we just want to reflect "custom configuration applied to limitador" for example ?

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+
+### The user wants more
+The user knows more about their deployment than we ever can.
+We need to allow the user to modify the configuration to suit their needs.
+When a user modifies the configuration, the kuadrant-operator should not reconcile back the changes.
+
+If the user changes are below the minimum spec that we state, this should be reflected in the status.
+If the user changes the configuration at all it should be reflected in the status. 
+However, when the spec is more than what we regard as minimum the status should be a highlight not warning or error.
+How this is reflected in the status of the Kuadrant CR is unknown currently.
```
maleck13, line 121: Yeah so I think perhaps (see above) we just aim to have a consistent status for when the configuration has been changed (whether below or above expectations) and a status for when it is as expected

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+    counterStorage: {} # lifts storage struct from the limitador CR.
+```
+
+## spec.resilience.authorization
+For `spec.resilience.authorization` we can make use of some features within the authorino CR, but will require updating to the authorino deployment and creation of PodDiruptionBudget CR.
+
+The authorino CR allows the setting of replicas `spec.replicas`.
+This should be set to 2. 
+The user should have the power to modify this number to what they want.
+If the number is less than the minimum we recommend (2), the kuadrant CR should report an error in the status.
```
maleck13, line 152: see above. don't think we need a different status for each just a "config changed status"

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+If the user uses a number greater than what we recommend, the kuadrant CR should report an information status.
+```yaml
+kind: Authorino
+metadata:
+  name: authorino
+spec:
+  replicas: 2
+```
+
+The deployment for authorino needs to be modified in the following was.
```
maleck13, line 162:
```suggestion
The deployment for authorino needs to be modified in the following ways.
```

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+          whenUnsatisfiable: ScheduleAnyway
+          labelSelector:
+            matchLabels:
+              limitador-resource: limitador
+```
+
+Limitador does require persisted storage.
+This can be configured in the limitador CR under `spec.storage`.
+If this section is missing from the limitador CR, and the `spec.resilience.rateLimiting`, the kuadrant CR needs to raise a warning in the status.
+The status is only a warning as it could be possible the user wants in memory counters.
```
maleck13, line 261: question for me though is when would they ever want to duplicate counters and have pretty unpredictable rate limiting depending on which limitador instance you hit? I don't see that as a use case or am I missing something? I think we could afford to be more opinionated. 
If you set ratelimiting to resiliant you **Must** specify a counter store?

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+
+## spec.resilience.counterStorage
+For `spec.resilience.counterStorage` the configuration will be added to the limitador CR under the `spec.storage` section.
+
+In the kuadrant CR this will be an object that matches the structure of the limitador [spec.storage](https://github.com/Kuadrant/limitador-operator/blob/626341d2aff5f6b8028317dc0e7d1bb27eb8d3d4/api/v1alpha1/limitador_types.go#L203-L212).
+Unlike the other resilience features, once configured the kuadrant-operator need to maintain the configuration within the limitador CR. 
+
+This will introduce an upgrade issue.
```
maleck13, line 269: hmm we can't break there installation during upgrade. We have a chicken and egg problem where this can't be added to the CR until after the upgrade has happened.
One option would be to introduce this change over a set number of releases. 
1) allow it to be there if already there. Add docs for the upgrade that strongly recommend adding the storage URL to the CR 
2) next version remove if present and again mention in upgrade docs that you need to have the storage URL in the CR before doing the upgrade.  WDYT

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,339 @@
...
+Even if the user has modified that configuration.
+
+This needs to be called out very clearly in the documentation.
+Why we do this is to simplify the reconciliation of configuration.
+
+### SRE reaction events
+From time to time SRE teams would need to change configurations for other to address issues.
+We should allow the user configure the deployment to a state that is we regard as not resilient.
+The one example that comes to mind is scaling replicas to zero for address some issue.
+The user should be allowed to that, but the Kuadrant CR should state the configuration is below expected spec.
```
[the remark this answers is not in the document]
Boomatang: In one sense, I agree, and in the other this is more for use. There has being a number time were I have seen an issue arise from testing (nightly, or RC), and the one of the first thing asked for is copy of the configure in question. I have never seen the alerts being asked for. I think both alerts, and good status are required. An alert will give information faster, but a status can give more detail.

rfcs/0000-resilient_deployment.md
Boomatang: I am not sure what the use case would be to have in memory counters with ratelimiting resiliency enable. So I am happy to have a Must specify a counter store if it is being used. I don't think it removes the ability of the user to use in memory counters, they would need to be explicit in their configuration to use them. 

It was mention somewhere else of using CEL to enforce this. That I think is a good idea.

rfcs/0000-resilient_deployment.md
Boomatang: I think what you mention in [Resilient Deployment
> Working in a iterative manner](https://kuadrant.zulipchat.com/#narrow/channel/497120-Resilient-Deployment/topic/Working.20in.20a.20iterative.20manner/with/512488926) can help here. Until we are happy that the feature done, ready for public use we keep the annotation in place. This will help with the chicken and egg problem allowing users to configure the features with out them being enabled, also they can make use of them.

rfcs/0000-resilient_deployment.md
Boomatang: I am not sure what we should do here. I am writing this RFC with the context that our existing [documentation](Resilient Deployment of data plane components) is correct and the [kuadrant/deployment](https://github.com/Kuadrant/deployment/tree/main) is also correct. 

I would not go changing how the Resource limits/requests are configured as part of this RFC. If there is question over what the correct configuration is I would remove them from the this RFC all together. At least until we can get our current recommendations updated.

rfcs/0000-resilient_deployment.md
Boomatang: I  found myself agreeing with this statement, but I can't see the ask. Is there an ask here?

rfcs/0000-resilient_deployment.md
Boomatang: That is good to call out.

rfcs/0000-resilient_deployment.md
maleck13: Ok I think copying what is in that repo works for this feature. As they only set the requests. You are correct, getting the right limits for the components is a different issue

rfcs/0000-resilient_deployment.md
maleck13: No :) just clarifying my own thoughts on it

rfcs/0000-resilient_deployment.md
```diff
@@ -0,0 +1,345 @@
...
+- [Limitador API](https://github.com/Kuadrant/limitador-operator/blob/main/api/v1alpha1/limitador_types.go#L203-L211)
+
+Configuration of the feature will be done in the Kuadrant CR.
+This is to make it simpler for the user.
+**API Design**
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+spec:
+  resilience:
```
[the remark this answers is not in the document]
maleck13: Given the storage config is needed. I see this as effectively a short cut to say the same thing as 

```
    authorization: True
    rateLimiting: True
    counterStorage: {"config"...} 
```

So when they set the top level enable value to true, we should treat it as though they have set authorization and rateLimiting to true and so expect counterStorage to be explicitly set also: Valid would be:

```
resilience:
      enable: true
      counterStorage: {"config"....}

```

One other note here. Authorino can do more than just `authorization` so I am not sure we should name this authorization and perhaps the short hand `auth` is a better option

Issue #118, opened 2025-04-08
RFC: AI & Gateway API policies


jasonmadigan:
Create a new RFC exploring potential new policies for:

- Token-based rate limiting
- Prompt guarding
- Guarding/risk checking for Inference completion

jasonmadigan:
@david-martin FYI

jasonmadigan:
Will create a draft in a couple of days, branch:

https://github.com/Kuadrant/architecture/compare/118-ai-policies?expand=1

PR #120, opened 2025-04-09
RFC: TokenRateLimitPolicy and future AI Policies

jasonmadigan:
Re: https://github.com/Kuadrant/architecture/issues/118

This is a new RFC to outline proposals to bring new policies to Kuadrant for LLM workloads


rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,491 @@
...
+  participant L as Limitador
+  participant MS as KServe Model Server
+  participant RPM as Response Processor (Usage Metrics Parser)
+
+  %% initial req
+  C->>HR: Incoming chat/completion request
+  HR->>E: Forward request
+
+  %% JWT authentication check via Authorino
+  E->>A: Validate JWT token & group access
```
alexsnaps, line 222: Minor nit here, but that's `WL` here, not `E`. Fundamentally, `WL` & `E` aren't really distinguishable from a user perspective. But in this particular context, I'd leave `E` out and "just" use `WL`, as that's where we can add/change behaviors as needed.

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,491 @@
...
+## New Policies
+
+### `LLMTokenRateLimitPolicy`
+
+A Kuadrant `LLMTokenRateLimitPolicy` is a custom resource provided by Kuadrant that targets Gateway API resources (`Gateway` and `HTTPRoute`) designed to allow users to define and enforce rate limiting rules to control token usage within OpenAI-style LLM provider APIs. It allows users to set and enforce token budget constraints for Gateways, but also for individual services exposed as HTTPRoutes. Per-user or per-group token rate limiting can be enforced based on JWT claims.
+
+
+### `LLMPromptRiskCheckPolicy`
+
+A Kuadrant `LLMPromptRiskCheckPolicy` is a custom resource provided by Kuadrant that targets Gateway API resources (`Gateway` and `HTTPRoute`), enabling users to define and enforce content safety rules with LLM prompts to detect and block sensitive prompts.  Prompt guards can be defined and enforced for both Gateways and individual HTTPRoutes.
```
maleck13, line 48: Would be interesting to get some feedback from a SME about what they and where they want things checked. In my naive view I can see the gateway being a good place to check the prompt. But I wonder about the response. Why would you not do that to be done before sending the response over the network?

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,491 @@
...
+
+### `LLMResponseRiskCheckPolicy`
```
maleck13, line 50: My instinct here as hinted at above is to start with the Prompt check and seek more opinion and feedback on the suitability of checking this at the gateway? Maybe it is still desirable to have a "last line of defence" type policy in place?

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,491 @@
...
+    name: my-llm-gateway
+  limit:
+    rate:
+      limit: 20000
+      window: 1d
+    predicate: 'request.auth.claims["kuadrant.io/groups"].split(",").exists(g, g == "free")' 
+    counter: auth.identity.userid
+---
+apiVersion: kuadrant.io/v1alpha1
+kind: LLMPromptRiskCheckPolicy
```
maleck13, line 119: I wonder could these be in the same API such as `LLMContentPolicy` but split underneath by the kuadrant operator as they are very similar? or do you expect the model being used to change based on the policy type?

```
spec:
  model: ....
  llmprompt:
    categories:
       ....
  llmresponse:
    categories:
response:
...
```

david-martin: They certainly could be in the same API.
My sole reason at the time for splitting them was usability.
That is, writing/configuring each part separately in a sizable block.

There may be a more concrete case for keeping them split.
Thoughts on if you want to apply a prompt check to a different group of users than a risk check?
Although not shown in this example, it could be useful to have a predicate for what users the policy applies to (based on auth headers etc..)
If combined into 1 policy, it would mean having multiple predicate fields.
Do we have a precedence for this with existing policies?

maleck13: yes RLP has that

```
  limits:
    "alice-limit":
      rates:

      - limit: 5
        window: 10s
      when:
      - predicate: "auth.identity.userid == 'alice'"
    "bob-limit":
      rates:
      - limit: 2
        window: 10s
      when:
      - predicate: "auth.identity.userid == 'bob'"
   ```

Comment (maleck13) [reply rfcs/0013-ai-policies.md:119]: so perhaps its a use case like

spec:
  model: ....
  llmprompt:
    - "under18":
       categories:
          ....
       when:
        - predicate: "auth.identity.age < 18"
      - "over18"
       categories:
          ....
       when:
        - predicate: "auth.identity.age >= 18" 
  llmresponse:
   -  "under18":

```

This is just sudo stuff but might be useful to think about?

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,486 @@
...
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+
+## `LLMTokenRateLimitPolicy`
+
+- Add a new resource, `LLMTokenRateLimitPolicy`, to be managed by the Kuadrant operator.
+- Either:
+  - Extend our existing `wasm-shim` to optionally amend the existing `actionSet` to optionally call both the guard filter and the token parsing filter implementation.
+  - Create a new `ext_proc` gRPC service for parsing OpenAI-style and usage metrics and adding these as well-known dynamic metadata, for use by Limitador
+- Extend the wasm-chim and `RateLimitPolicy` to give a means to specify an increment (currently, [hard-coded](https://github.com/Kuadrant/wasm-shim/blob/main/src/service/rate_limit.rs#L18) to `1`)
```
maleck13, line 195:
```suggestion
- Extend the wasm-shim and `RateLimitPolicy` to give a means to specify an increment (currently, [hard-coded](https://github.com/Kuadrant/wasm-shim/blob/main/src/service/rate_limit.rs#L18) to `1`)
```

rfcs/0013-ai-policies.md
alexsnaps: nit: `request.model` seem to be wanting to "enhance" the `request` struct with arbitrary fields. I'd advise against that. When is that `model` field present? I don't want users to ask themselves these questions. I understand this is just for illustration purposes, but nonetheless raises an interesting question: where would "additional policies" append data to the "well-known attributes"?

maleck13: Yes where we append this data is an important consideration. Perhaps we need a new namespace for AI metadata. ```ai.model```

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,486 @@
...
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+
+## `LLMTokenRateLimitPolicy`
+
+- Add a new resource, `LLMTokenRateLimitPolicy`, to be managed by the Kuadrant operator.
+- Either:
+  - Extend our existing `wasm-shim` to optionally amend the existing `actionSet` to optionally call both the guard filter and the token parsing filter implementation.
+  - Create a new `ext_proc` gRPC service for parsing OpenAI-style and usage metrics and adding these as well-known dynamic metadata, for use by Limitador
+- Extend the wasm-shim and `RateLimitPolicy` to give a means to specify an increment (currently, [hard-coded](https://github.com/Kuadrant/wasm-shim/blob/main/src/service/rate_limit.rs#L18) to `1`)
```
maleck13, line 195: I don't know if this is needed in the RLP api? it seems a very special use case. I wonder instead if it could be made some form of internal config not exposed to the user at this point?

jasonmadigan: that is an option yes: we either inc the counter in a custom fashion in something in this filter chain, or we extend RLP somehow to support. if there's general utility in RLP, I guess that route may be more preferable

slight worry about having another mechanism other than limitador doing it - would end up needing to re-implment/copy a bunch of existing machinery? unsure

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,486 @@
...
+  end
+
+  %% final inference response: deliver back to client
+  WL->>HR: Forward final response
+  HR->>C: Deliver final LLM response
+```
+
+### Parsing OpenAI-style usage metrics
+
+OpenAI-style usage metrics for both completion and response APIs generally have a `usage` object, with values for `prompt_tokens` (token count for the initial prompt), `completion_tokens` (tokens generated by the model in response) and `total_tokens` (prompt + response tokens count).
```
maleck13, line 305: is there one of these that is more popular than the other? I am wondering does it make sense to support just one for now and expand beyond that in the future?

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,486 @@
...
+
+```json
+{"stream_options: {"include_usage": true}}
+```
+
+In streamed responses, `usage` metrics are included in either the second-to-last or last chunk in a response.
+
+Given the permutations, this will add some extra complexity to how we parse usage metrics. There is a basic Golang example of an `ext_proc` that can parse these metrics (non-streamed responses) here: https://github.com/jasonmadigan/token-ext-proc
+
+We will also want to support llama-stack style responses. Inference chat-completion with llama-stack offers the option for configurable (JSON-schema) guideded `response_format`. This may hint that we'll want to offer some customisation in terms of where to look for metrics, (probably CEL, or JQ-style querying?).
```
maleck13, line 317: beyond these tokens are there other reasons a user may want to pull something from the body? I think I would start with inferring and looking for these specific values rather than surfacing them into the API at this stage?

maleck13: As in I would prefer the user to have to hint at the response types to expect and we then use that to decide what values to pull out rather than open up the entire request body to the user to pull values out of.

rfcs/0013-ai-policies.md
jasonmadigan: assume this is re: Chat completion API vs Responses API

The chat completion API is more universally supported, the Responses API is newer, but is designed to work with more use-cases (agentic use-cases, as well as support for "reasoning/show my thoughts" response streaming)

maleck13 (commented):
This is a really great start to some cool features. I think we still need to nail down how we want to do the request filtering (seems to be leaning towards the WASM shim) and also whether we need to expose certain options to the user or not

rfcs/0013-ai-policies.md
jasonmadigan: Reason I think we may want to offer some sort of API here is that although lots of runtimes offer openai-style completion APIs, in the ones I've looked at there are some small differences (i.e. where in the JSON response usage metrics are) which could break the policy. We could have some prebuilt "variants" to make the APIs look nice though (perhaps starting with one for llama-stack and one for openai-style) - these variants would come with builtin selectors on what attributes to pluck to get our usage metrics (if that makes sense)

maleck13: Yeah like we could for now call out those as the supported options. Rather than jumping straight to giving the power to the user which we might regret and not be able to take back. That said we are talking about Alpha apis so easier to take back then other places

jasonmadigan: that makes sense. I suppose internally we'll probably use selectors, and then if we decide later, we can expose those to end users

rfcs/0013-ai-policies.md
maleck13: hmm I might be misunderstanding. I was instead thinking of using the `hits_addend` and setting it dynamically if it is an AI interaction. We need @eguzki or @alexsnaps here interested to know what there thoughts are on how to send a custom increment.

jasonmadigan:
One other potential policy which may emerge here, depending on how [this PoC progresses](https://github.com/Kuadrant/kuadrant-operator/issues/1279), is a `SemanticCachingPolicy` for short-circuiting (well, partially, still need to do embedding) expensive LLM calls if we see similar prompts

rfcs/0013-ai-policies.md
alexsnaps: While this isn't implemented afaik, [well-known attributes](https://github.com/Kuadrant/architecture/blob/main/rfcs/0002-well-known-attributes.md#rate-limit-attributes) were meant to support this. So one way would be to have `ratelimit.hits_addend` (while defaulting to `1`) be mutable by "upstream" actions. So that a policy could set it to some arbitrary value _before_ the request to limitador is made.

alexsnaps: and, yes! 🙌 , `service.ext_proc.v3.ProcessingResponse` support [dynamic metadata](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ext_proc/v3/external_processor.proto#envoy-v3-api-field-service-ext-proc-v3-processingresponse-dynamic-metadata) to have data flow between envoy and these processes. If wasm-shim dispatches the call, then no worries in using that for that. If envoy does, we have to check what happens to it and how/if we can read them back properly from wasm tho (should be fine™ - t&c apply)

rfcs/0013-ai-policies.md
david-martin: Depending on how you use it, it could be a last line of defence.

It could also be a targeted response check based on a specific group of users for example, since you have access to the users auth context. That capability may not be easily available in the model serving runtime without custom logic.

rfcs/0013-ai-policies.md
david-martin: It may not be possible to avoid network traffic, like in the case of streaming the response.
You may end up streaming/chunking the response and checking it all, or in parts.

rfcs/0013-ai-policies.md
```diff
@@ -38,14 +38,14 @@ Using the [Inference Platform Admin](https://github.com/kubernetes-sigs/gateway-
...
 
 
-### `LLMPromptRiskCheckPolicy`
+### `PromptGuardPolicy`
 
-A Kuadrant `LLMPromptRiskCheckPolicy` is a custom resource provided by Kuadrant that targets Gateway API resources (`Gateway` and `HTTPRoute`), enabling users to define and enforce content safety rules with LLM prompts to detect and block sensitive prompts.  Prompt guards can be defined and enforced for both Gateways and individual HTTPRoutes.
+A Kuadrant `PromptGuardPolicy` is a custom resource provided by Kuadrant that targets Gateway API resources (`Gateway` and `HTTPRoute`), enabling users to define and enforce content safety rules with LLM prompts to detect and block sensitive prompts.  Prompt guards can be defined and enforced for both Gateways and individual HTTPRoutes.
```
jasonmadigan, line 48: +1 on these policy name changes btw

rfcs/0013-ai-policies.md
```diff
@@ -176,31 +176,29 @@ spec:
...
 
-- Its interaction with other features is clear.
-- It is reasonably clear how the feature would be implemented.
-- How error would be reported to the users.
-- Corner cases are dissected by example.
+- Add a new resource, `TokenRateLimitPolicy`, to be managed by the Kuadrant operator.
+- Extend the `wasm-shim` to:
+  - Add a new `action` to the `actionSet` types to execute the token rate limiting logic on the request and response.
+  - Implement the rate limit checking logic as the request body is processed. The request body has to be parsed to know which model is being targeted.
+    - Initial descriptors would include the request path, user id (if available) and the requested model.
```
jasonmadigan, line 185: descriptors are available to later filters right? i.e. if/when we have a responseguardpolicy, ext_proc would have access to model / token usage metrics?

rfcs/0013-ai-policies.md
```diff
@@ -176,31 +176,29 @@ spec:
...
-- Look at ways to support a "two-phase" approach to rate-limiting: a standard, normal rate limit enforcement prior to hitting the model server (responding early if limited), and (if not limited) one after based on one to increment counters by a custom increment (after parsing usage metrics)
+- Add a new resource, `PromptGuardPolicy`, to be managed by the Kuadrant operator.
+- Extend the `wasm-shim`, ammending the `actionSet` to optionally call a prompt guard filter implementation in an `ext-proc` service.
+  - Only the request body phase needs to be handled by the `ext-proc` service.
+- The `ext-proc` service contains all logic to parse the prompt from the request body and pass it to a configured guard model.
```
jasonmadigan, line 196: makes sense, let's build it an prove the hypothesis

rfcs/0013-ai-policies.md
david-martin: That makes sense.
In GIE, there is 'body based routing' or [bbr component](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/pkg/bbr) that parses the model and puts it in a `X-Gateway-Model-Name` header.
We could do the same wherever the model is needed first, perhaps in the wasm-shim.
The user metadata would be whatever is populated by the auth actionset/authorino.
The request path should already be available in the `request.url_path` [envoy attribute](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#request-attributes).

Token usage metrics will be parsed in the wasm-shim, with the proposed approach.
I think this is a good way to progress an implementation while also discussing a provision upstream for if token usage is parsed in an external service, and having that populated in a header or attribute for subsequent filters to use.

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,481 @@
...
+- As an Inference Workload Owner, I want to apply guards against different risky content categories to different user groups based on business requirements.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## New Policies
+
+### `TokenRateLimitPolicy`
+
+A Kuadrant `TokenRateLimitPolicy` is a custom resource provided by Kuadrant that targets Gateway API resources (`Gateway` and `HTTPRoute`) designed to allow users to define and enforce rate limiting rules to control token usage within OpenAI-style LLM provider APIs. It allows users to set and enforce token budget constraints for Gateways, but also for individual services exposed as HTTPRoutes. Per-user or per-group token rate limiting can be enforced based on JWT claims.
```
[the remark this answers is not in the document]
david-martin: I think it's fine to go with what's familiar at first, like the counters expression example, as it can deliver functionality and get feedback sooner.

There is value in a higher level abstraction if it makes it easier for users.
For example, having a jwt specific config block in the TRLP CRD, which results in an AuthPolicy/AuthConfig being created under the hood. This would also avoid having to validate a cel expression and ensuring what's in the cel expression has a dependent AuthPolicy being enforced.

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,541 @@
...
+apiVersion: kuadrant.io/v1alpha1
+kind: TokenRateLimitPolicy
+metadata:
+  name: token-limit-free
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: my-llm-gateway
+  limit:
```
jasonmadigan, line 111: an earlier version of this more closely mirrored `RateLimitPolicy`, in that multiple `limits` defined could be . Wondering why we diverged from that - simplicity? one for @david-martin later

cc @eguzki

jasonmadigan:
@david-martin pushed that commit- moves the other proposed policies to future possibilities

eguzki (approved):
LGTM

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,542 @@
...
+    kind: Gateway
+    name: my-llm-gateway
+  rules:
+    authorization:
+      "gold-users":
+        opa:
+          rego: |
+            groups := split(object.get(input.auth.identity.metadata.annotations, "kuadrant.io/groups", ""), ",")
+            allow { groups[_] == "gold" }
+---
```
maleck13, line 73: I get this is just an example but multiple authpolicies here seems odd? trying to recall if multiple policies targeting the same resource even work

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,542 @@
...
+- The order of actions matters here, as usage metrics are flushed as part of the body of LLM responses (either complete responses, or when streamed). Some additional notes on our existing filters, including our "internal to WASM" http filter chain, in this thread: https://kubernetes.slack.com/archives/C05J0D0V525/p1744098001098719. A flow diagram below attempts to capture this flow at a high level.
+- Look at ways to avoid 2 requests to limitador per single request to a model. This is not ideal to have a limit check and counter increment happen separately due to scaling concerns. However, this approach is sufficient for an initial implementation.
+- A new action type is not being considered. The WASM module will only initiate a ShouldRateLimit gRPC call to Limitador when all associated CEL expressions (namely `predicates` and `data`) can be evaluated.
+- If any of the CEL expression references the `requestBodyJSON()` CEL function, the gRPC request will be triggered after the downstream request body has been parsed.
```
maleck13, line 220: duplicated?

rfcs/0013-ai-policies.md
```diff
@@ -0,0 +1,542 @@
...
+- If any of the CEL expression references the `responseBodyJSON()` CEL function, the gRPC request will be triggered after the upstream response body has been parsed.
+- The order of actions is important and will be enforced:
+  - If one action requires evaluation of the `requestBodyJSON()` and a subsequent action can be performed during the request headers phase, both actions will be executed during the request body phase.
+  - If one action requires evaluation of the `responseBodyJSON()` and a subsequent action can be performed during any of the previous request phases, both actions will be executed during the response body phase.
```
maleck13, line 224: are both of these needed seems the second one is saying the same thing?

maleck13 (approved):
Couple of small comments otherwise LGTM

rfcs/0013-ai-policies.md
jasonmadigan: i'll change this - was written before impl, the examples from the tutorials do similar with a single policy

rfcs/0013-ai-policies.md
jasonmadigan: resolved

PR #122, opened 2025-04-10
[WIP] First draft of DNS Publishing RFC

philbrookes:


thomasmaas:
Wonder if we're overloading 1 field of the CRD with a lot of information? But Maybe it's ok and the possible complexities here can be simplified with a set of meta policies that cover most common publishing use cases.

maleck13 (commented):
hmm. Thinking about this, how would a user achieve something like this right now. 
Assuming they had a policy per listener, they would see a certain state they could delete the policy manually which would cause the DNSRecord to be removed and recreate the policy manually to cause the records to be re-published.

I am thinking that perhaps the DNS operator should be responsible for reflecting state and then the DNSPolicy makes the DNSRecord available for querying on a per listener basis and then it is up to the Kuadrant Operator to decide to delete, create or unpublish a record. So it would be the Kuadrant Operator that would evaluate the CEL expression

```
publishStrategy:
  rule: listeners['api'].dnsrecord.status.healthy == 'true'
```
I think as Kuadrant is the owner of the records and has the context of the records, the gateways etc. The policy should decide when something is published or not and I don't see a need to push that into the DNSOperator.

rfcs/0013-dns-publishing-strategy.md
```diff
@@ -0,0 +1,97 @@
...
+```yaml
+unpublishRules:
+  <rule implemented using Common Expression Language>
+  <rule implemented using Common Expression Language>
+  <rule implemented using Common Expression Language>
+```
+### Kuadrant and DNS Operator operator changes
+
+### Health check probe creation
+The health checks are currently created by the DNS Operator, this will instead be moved to the Kuadrant-operator and the health checks will be owned by the DNS Policy that defined them, they will no longer have a failureThreshold as this can be defined in the unpublishRule, this section will no longer be copied to the DNS Record. They will continue to be reconciled by the DNS Operator.
```
maleck13, line 48: We can't remove the failureThreshold. That will be a breaking change to the API. In my opinion we could map failure theshold to a predefined cel predicate so as to maintain back compat but keep the impl the same for both

maleck13: By reconciled by the DNS Opertor I assume that means the health checks will still be performed by the DNS operator

philbrookes: updated to apply these changes.

philbrookes:
> > One thought that occurred to me, is that we have an implicit "publish" condition (health checks are passing and there is a HTTPRoute assigned to the host) and are describing explicit unpublish conditions. So I am wondering should we expand this to cover both: (sudo code incoming)
> > ```yaml
> > spec:
> >    unpublish:
> >       when: 
> >            - 'target.status.attachedRoute == 0'
> >    publish:
> >       when: 
> >          - "target.status.attachedRoutes > 0 && self.healthy == true && self.metadata.annotations["some.annotation"] == 'failOver'"
> > ```
> 
> my inclination, for this example, would be to update the unpublish rule:
> 
> ```yaml
> spec:
>    unpublish:
>          - "target.status.attachedRoutes <= 0 || self.healthy == false || self.metadata.annotations["some.annotation"] != 'failOver"
> ```

more conversation on this, here: https://kuadrant.zulipchat.com/#narrow/channel/493518-DNS/topic/publishing.20strategy/near/523039789

maleck13:
@philbrookes do you see this as still relevant?

PR #126, opened 2025-06-16
allow route defined hostnames for dns and tls

maleck13:
related to # https://github.com/Kuadrant/kuadrant-operator/issues/1386

Defines an API for controlling behaviour of DNSRecords and TLS certs when using a wildcard hostname

### DIFF
--- a/rfcs/route-defined-host-names.md
+++ b/rfcs/route-defined-host-names.md
@@ -1,0 +1,60 @@
+# RFC Template
+
+- HTTPRoute defined Host Names
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [GH Issue](https://github.com/Kuadrant/kuadrant-operator/issues/1386)
+
+# Summary
+[summary]: #summary
+
+When attaching a DNSPolicy or TLSPolicy to a Gateway with a listener that has a wildcard hostname e.g. *.app.com defined, we default creating a wildcard DNS record to match and a wildcard TLS certficiate to secure communication. A wildcard DNS name and TLS certificate may not always be desirable espcially if you want to take advantage of health checks. Some users might want to more tightly control which hostnames are served and defined in DNS and have certificates. While this could be achieved by adding specific listeners to a gateway, this could become quite cumbersome for platform engineers that want to enable development teams to self serve or if a large number of unique hostnames are required. Additonally health checks do not operate against wildcard hosts as they do not provide any real value given there could be many workloads behind a single wildcard.  
+
+# Motivation
+[motivation]: #motivation
+
+- Add better control over what happens with a wildcard hostname defined at the gateway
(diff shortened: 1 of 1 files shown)


rfcs/route-defined-host-names.md
```diff
@@ -0,0 +1,65 @@
...
+- Enable health checks to work when using a wildcard listener.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When using DNSPolicy or TLSPolicy you will now be able to define whether you want a wildcard certificate and DNSRecord or if you want to delegate the hostname definition to the attached HTTPRoutes. To delegate you will need to set `AllowWildcardHosts` to `false`. When delegated, the DNSPolicy will setup a new DNSRecord for each unique hostname defined under a listener with a wildcard hostname and TLSPolicy will setup a new multi-hostname certificate . 
+
+We will achieve this by adding an `AllowWildcardHosts` to the DNSPolicy and TLSPolicy. if this is set to false, no wildcard records will be created and no wildcard TLS certificates will be created regardless of the hostnames specified. An empty listener hostname is considered a wildcard.  If `AllowWildcardHosts` is false and a wildcard is defined in a HTTPRoute, the DNS and TLS policy will report that no record or tls certificate has been created for that listener / httproute via the status of the policy. So effectively the route will not be exposed via DNS.
```
philbrookes, line 24: I'm not loving this field name. As what it is trying to describe is not whether or not wildcard hosts are allowed, but rather whether or not the Kuadrant Operator will find all created routes that have attached to a wildcard host. 

I think it makes more sense to call this: `HostsFromRoutes`.
- false: (as current) Kuadrant Operator reconciles DNS Records from listeners in targeted gateway.
- true: Kuadrant Operator reconciles DNS Records from routes attached to the targeted gateway.

rfcs/route-defined-host-names.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+A new property `AllowWildcardHosts (bool)`  will be added to both TLS and DNS policy.  It will default to true, as this is the current behaviour. When set to false, it will take the following actions:
```
philbrookes, line 31: Have you considered this value being changed after reconciling for some time? In particular if there is a wildcard listener present.

Changing from routes to wildcard would result in a mass delete of DNS Records and then a subsequent create of the wildcard, and vice versa (presenting a possible NXDomain response window).

rfcs/route-defined-host-names.md
maleck13: I have actually moved to something closer to what your are suggesting locally when investigating this, It is now ```useHostsFromRoutes```

rfcs/route-defined-host-names.md
maleck13: Will update with a little detail. But the delete only happens once the new record is ready which should reduce / remove any issue

PR #127, opened 2025-07-01
rfc: DNS Cluster Aware Delegation

mikenairn:
Proposal to add functionality to delegate the processing of a DNSRecord to a designated cluster or clusters in a multi cluster environment.

### DIFF
--- a/rfcs/0000-dns-cluster-aware-delegation.md
+++ b/rfcs/0000-dns-cluster-aware-delegation.md
@@ -1,0 +1,160 @@
+# DNS Cluster Aware Delegation
+
+- Feature Name: `dns_cluster_aware_delegation`
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Proposal to add functionality to delegate the processing of a DNSRecord to a designated cluster or clusters in a multi cluster environment. 
+
+# Motivation
+[motivation]: #motivation
+
+Multi cluster DNS is currently achieved by using the eventual provider DNS service (AWS Route etc ..) as a store for ownership metadata using specially created TXT records, and as a central API service that all clusters can communicate with.
+Each cluster processes its own DNSRecords, becoming aware of other DNSRecords contributing to the same set of endpoints via this centrally stored data, in turn allowing it to correctly translate the DNSRecord endpoints into an appropriate API operation (Create/Update/Delete) and get to the desired state. 
+
+In some cases, such as our current CoreDNS solution, there is no central off cluster API service that can be configured on all clusters and as such a different approach to multiple cluster discovery and reconciliation is required.
+
+* Add a provider agnostic alternative for multi cluster dns support while still maintaining current behaviour for existing providers.
+* Reduce the number of required CoreDNS instances to one, but still allow multiple if required, redundancy etc...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Delegating Authority to Publish
+
+Delegating authority is a concept whereby you can optionally specify that you do not want each DNSPolicy to publish its endpoints directly to a provider, but instead have a central authoritative DNSRecord be created that will take care of publishing all endpoints for a particular set of records.
+
+This feature is controlled by the `authorityDelegation` field in the DNSPolicy spec, which has a single required field `role` which must be one of `primary` or `remote`.
+
+- `spec.authorityDelegation` when set signals that records produced by this policy will have their endpoints published via an authoritative record
+- `spec.authorityDelegation.role` signals what role this particular policy is taking, it is expected there should be at least one `primary`
+
+### Primary Role
+
+There should always be at least one primary, and the primary is in charge of ensuring that the authoritative DNSRecord that all publishing requests are delegated to is created.
+The authoritative DNSRecord is created in the same namespace as the original DNSRecord created by the DNSPolicy and at a minimum will contain the original DNSRecords endpoints.  
+
+### Remote Role
+
+A remote will only have its endpoints added to an authoritative DNSRecord that already exists in the current namespace and does not initiate the creation of the authoritative DNSRecord.
+If an authoritative DNSRecord does not exist, an error will be reported until such time as one does exist.
+
+**Example 1.** DNSPolicy using authority delegation with `primary` role
+```yaml
+apiVersion: kuadrant.io/v1
+kind: DNSPolicy
+metadata:
+  name: prod-web-coredns
+spec:
+  authorityDelegation:
+    role: primary
+  targetRef:
+    name: prod-web-istio-coredns
+    group: gateway.networking.k8s.io
+    kind: Gateway
+  providerRefs:
+    - name: dns-provider-credentials-coredns
+```
+
+**Example 2.** DNSPolicy using authority delegation with `remote` role
+```yaml
+apiVersion: kuadrant.io/v1
+kind: DNSPolicy
+metadata:
+  name: prod-web-coredns
+spec:
+  authorityDelegation:
+    role: remote
+  targetRef:
+    name: prod-web-istio-coredns
+    group: gateway.networking.k8s.io
+    kind: Gateway
+  providerRefs:
+    - name: dns-provider-credentials-coredns
+```
+
+If the above example policies were created, three DNSRecord resources would exist in the target namespace, with a single DNSRecord resource (the authoritative) appearing as the owner of all endpoints in the target dns provider.
+
+Note: While this concept is intended to work with the multi cluster concept described below, it is also possible to use this on a single cluster to combine several records together. 
+This might be desirable if you have many records and want to reduce the number of API calls to a provider and also the number of owner TXT records that need to be managed.
+
+[//]: # (mnairn: Add a diagram or something to visualise the records being produced  )
+
+## Multi Cluster
+
+The DNS reconciler can be made cluster aware allowing it to process DNSRecord resources on other clusters as well as its own.
+
+Making the controller aware of other clusters is done via specially labeled secret resources containing kubeconfig data that exist in the `kuadrant-system` namespace.
+
+The kubeconfig data should be generated from an appropriately scoped service account that exists on the target cluster.
+
+When a cluster secret is created, the DNS controller on that cluster will automatically detect it and start watching for DNSRecord resources on that external cluster as well as its own.
+
+**Example** Secret containing kubeconfig data
+```yaml
+apiVersion: v1
+kind: Secret
+data:
+  kubeconfig: <kubeconfig data>
+metadata:
+  labels:
+    kuadrant.io/multicluster-kubeconfig: "true"
+  name: kind-kuadrant-local-2
+  namespace: kuadrant-system
+type: Opaque
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## CRD Provider
+
+A new provider implementation that allows a DNSRecord resource to act as a zone record will be created. 
+Endpoints from the source are injected into the target "zone" record using the same plan logic as all other providers. 
+A zone record could be getting updated by many DNSRecords just like any other providers zone.
+
+## Authority Delegation
+
+A new API on the DNSRecord that allows a DNSRecord (and DNSPolicy) to signal that it is delegating authority of publishing to another DNSRecord. 
+Internally uses the CRD provider described above to do this. 
+Assigns a role to the delegation, one of "primary" or "remote" to signal what this particular records role is. 
+Primary records will update/create the authoritative record on the same cluster and supply the provider credentials, remote records only update existing zone records.
+
+## Multicluster
+
+Updates the dns operator to allow watches for DNSRecords on different clusters, ultimately allowing a single controller to reconcile DNSRecords from multiple clusters. 
+Clusters are configured via Secrets containing kubeconfig data and labelled as cluster secrets.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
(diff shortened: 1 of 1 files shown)


rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+## Delegating Authority to Publish
+
+Delegating authority is a concept whereby you can optionally specify that you do not want each DNSPolicy to publish its endpoints directly to a provider, but instead have a central authoritative DNSRecord be created that will take care of publishing all endpoints for a particular set of records.
+
+This feature is controlled by the `authorityDelegation` field in the DNSPolicy spec, which has a single required field `role` which must be one of `primary` or `remote`.
+
+- `spec.authorityDelegation` when set signals that records produced by this policy will have their endpoints published via an authoritative record
+- `spec.authorityDelegation.role` signals what role this particular policy is taking, it is expected there should be at least one `primary`
```
philbrookes, line 34: is there any way for the controllers to determine that no primary exists and set a status on the policy?

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Delegating Authority to Publish
+
+Delegating authority is a concept whereby you can optionally specify that you do not want each DNSPolicy to publish its endpoints directly to a provider, but instead have a central authoritative DNSRecord be created that will take care of publishing all endpoints for a particular set of records.
+
+This feature is controlled by the `authorityDelegation` field in the DNSPolicy spec, which has a single required field `role` which must be one of `primary` or `remote`.
+
+- `spec.authorityDelegation` when set signals that records produced by this policy will have their endpoints published via an authoritative record
```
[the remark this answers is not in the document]
mikenairn: An empty `spec.authorityDelegation` will just work as it does today where the DNSPolicy creates a DNSRecord and that record publishes it's endpoints directly to the provider and becomes an owner of those endpoints. 

I think the names "primary" and "remote" are maybe a bit misleading and therefore probably wrong in this context. What you say makes sense, current behaviour is "primary", but that isn't really what is being proposed since being a primary is telling it to create this secondary authoritative record and delegating the responsibility of publishing to it.

A nicer way might be to just make the existing "primary" DNSRecord the authoritative record, i.e. no secondary one, but the kuadrant operator owns those and the dns operator shouldn't really be messing with the spec of it.

rfcs/0000-dns-cluster-aware-delegation.md
mikenairn: Yeah, well i think at a minimum we would want that and i did mention something [here](https://github.com/Kuadrant/architecture/pull/127/files#diff-4c24f82bc117f776d24141a6246aff55e71aaa9d5bf560c30dda501c3ca625fbR44) which is effectively the same, no primary no authoritative

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+If an authoritative DNSRecord does not exist, an error will be reported until such time as one does exist.
+
+**Example 1.** DNSPolicy using authority delegation with `primary` role
+```yaml
+apiVersion: kuadrant.io/v1
+kind: DNSPolicy
+metadata:
+  name: prod-web-coredns
+spec:
+  authorityDelegation:
```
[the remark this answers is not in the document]
mikenairn: Yeah, maybe its possible to describe this in one field at the top level of the spec. I see three paths currently:

1. I am a DNSRecord publishing my endpoints directly to the provider (Current behaviour).
2. I am a DNSRecord providing endpoints and additional config (credentials) to another DNSRecord responsible for publishing mine and other endpoints to the provider.
2. I am a DNSRecord providing endpoints only to another DNSRecord responsible for publishing mine and other endpoints to the provider.

So maybe three roles with the one doing the publishing ideally having the same role in each case.

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+A zone record could be getting updated by many DNSRecords just like any other providers zone.
+
+## Authority Delegation
+
+A new API on the DNSRecord that allows a DNSRecord (and DNSPolicy) to signal that it is delegating authority of publishing to another DNSRecord. 
+Internally uses the CRD provider described above to do this. 
+Assigns a role to the delegation, one of "primary" or "remote" to signal what this particular records role is. 
+Primary records will update/create the authoritative record on the same cluster and supply the provider credentials, remote records only update existing zone records.
+
+## Multicluster
```
maleck13, line 127: I think we should add details about the permissions the primary will need in the remote clusters

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+* Adds another path through the code that isn't strictly necessary for the majority of providers.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The initial idea here was to look at a way to "push" records from multiple "remote" clusters to a central "primary" cluster that would be in charge of combining those records into a single "authoritative" record and ultimately in charge of publishing that record to the provider (CoreDNS). 
+This would require all clusters to have credentials capable of writing to the "primary" cluster.
+
+However, it appears to be a more common pattern in kubernetes projects that deal with multi cluster communication to have the "primary" clusters read from multiple "remote" clusters. 
+This would also only require the "primary" to have read access, or very limited write (status only) access to each cluster.
```
maleck13, line 144: I see it is called out here on the perms. But I think we should put it under the reference level stuff

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+    kuadrant.io/multicluster-kubeconfig: "true"
+  name: kind-kuadrant-local-2
+  namespace: kuadrant-system
+type: Opaque
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## CRD Provider
```
maleck13, line 114: It would be good to see an example CRD provider

maleck13 (commented):
I think the bones of it is here. I think we should call out in a bit more detail

- how conflicts from a secondary DNSRecord would be handled 
- how deletion and clean up is handled
- The minimum perms the primary needs to make this work
- What a CRD provider looks like

I also wondered if it would make things easier to reason about if the DNS Operator/Kuadrant had a flag set from a configmap that would put it into "secondary or primary" mode. You could then have the kuadrant resource accept configuration of 
```
dns:
  mode: secondary
```
that would trigger it to update the configmap mounted into the dns operator. This leaves an open question around providers though

rfcs/0000-dns-cluster-aware-delegation.md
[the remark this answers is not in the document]
mikenairn: out of date, updated to use controller flags for "mode", options being "default, "primary" or "remote"

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Primary/Remote Configuration.** Multi cluster using primary/remote modes and cluster aware controller
+
+![img.png](./0000-dns-cluster-aware-delegation-assets/multi-custer-primary-remote-cluster-aware.png)
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## CRD Provider
```
mikenairn, line 98: I'm thinking this might be better called "dnsrecord" provider since that's what it actually targets, CRD might be suggesting it could be anything.

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,170 @@
...
+* Add a provider agnostic alternative for multi cluster dns support while still maintaining current behaviour for existing providers.
+* Reduce the number of required CoreDNS instances to one, but still allow multiple if required, redundancy etc...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## DNS Operator modes
+
+The DNS Operator can be configured to work in different modes specified via a controller flag `mode` which must be one of `default`, `primary` or `remote` i.e. `--mode=[default|primary|remote]`.
+To allow for customisation of this mode via an OLM deployment of the kuadrant-operator, the mode will also be configurable via an environment variable `DNS_MODE`.
```
maleck13, line 30: I think it is fine to have the env var, but we should still consider surfacing the dns mod via the kuadrant CR

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,170 @@
...
+
+DNSRecords will be reconciled and published to the provider by this instance through an authoritative DNSRecord which is created on behalf of the primary DNSRecord that all publishing requests are delegated to.
+DNSRecords on these clusters must have access to valid provider credentials referenced in the spec.
+The authoritative DNSRecord is created in the same namespace as the original DNSRecord (created by the DNSPolicy or otherwise) and at a minimum will contain the primary DNSRecords endpoints.
+The remaining specification of the authoritative DNSRecord is synced from the original primary DNSRecord (rootHost, providerRef etc..) 
+This mode should be selected for multi cluster scenarios where a designated "control plane" cluster(s) are being configured with elevated privileges to access external DNS providers.
+
+### Remote
+
+DNSRecords will be reconciled by this instance, but only to maintain a finalizer on the resource.
```
maleck13, line 50: it will maintain finalizer and also health checks

rfcs/0000-dns-cluster-aware-delegation.md
maleck13: I was thinking if with kuadrant we provided a configmap that the DNSOpertor used for env vars. If installing the dns operator yourself you can just set the env var, but with kuadrant we provide a configmap that the kuadrant operator could then use to configure the DNS Operator?

maleck13: You can do it via a subscription, but I think a configmap is a more generic approach as you say. No it doesn't need to be done as part of this work but is something we should likely capture

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -0,0 +1,160 @@
...
+  namespace: kuadrant-system
+type: Opaque
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## CRD Provider
+
+A new provider implementation that allows a DNSRecord resource to act as a zone record will be created.
```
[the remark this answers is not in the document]
mikenairn: I updated the doc around this. A primary cluster will create these records with a label such as `kuadrant.io/primary-authoritative-zone-record=$rootHost`. This allows you to list them it needs be, but also it will allow the dnsrecord provider to be configured to look for an exact match of the rootHost when kuadrant is managing the zone record as discussed on our last call, but not limiting the dnsrecord provider to being only able to work that way.

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -34,33 +34,36 @@ The mode will determine the behaviour of the reconciliation of all DNSRecords th
...
-DNSRecords on these clusters must have access to valid provider credentials referenced in the spec.
-The authoritative DNSRecord is created in the same namespace as the original DNSRecord (created by the DNSPolicy or otherwise) and at a minimum will contain the primary DNSRecords endpoints.
-The remaining specification of the authoritative DNSRecord is synced from the original primary DNSRecord (rootHost, providerRef etc..) 
-This mode should be selected for multi cluster scenarios where a designated "control plane" cluster(s) are being configured with elevated privileges to access external DNS providers.
+DNSRecords will be reconciled and published to the provider through an authoritative zone DNSRecord which is created on behalf of the primary DNSRecord with all publishing requests delegated to it.
+The dns operator will ensure the existence of the authoritative zone DNSRecord, with:
+  * the same `metdata.namespace` as the current record
+  * a `metadata.name` generated based on the current records `spec.rootHost` i.e. `foo-example-com_authoritative_zone`
```
maleck13, line 45: nit: I know just an example but prob would keep the suffix as short as possible as there is a max length to consider

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -79,35 +82,70 @@ type: Opaque
...
-ToDo Expectations of layout of namespaces and names of DNSRecords to match across clusters
+ToDo mnairn: Expectations of layout of namespaces and names of DNSRecords to match across clusters
+
+## Default Provider Secret
+
+A new concept of "default provider secret" will be introduced that removes the need for a `providerRef` to be specified on every DNSRecord and DNSPolicy.
+When a DNSPolicy is created with no `providerRef` the resulting DNSRecord during initial reconciliation will search for a default provider secret in the current namespace to use instead.
+The DNSRecord status will be updated to include the selected provider secret as a `providerRef` in the status.
+A default provider secret is any provider secret that has been appropriately labelled with the kuadrant default provider label (kuadrant.io/default-provider=true)
+Only one default provider can exist in any namespace and more than one labelled as such will result in an error (mnairn: I have actually no idea where????)
```
maleck13, line 93: It should populate onto the status of the DNSRecord IMO which will then also show on the DNSPolicy

rfcs/0000-dns-cluster-aware-delegation.md
```diff
@@ -79,35 +82,70 @@ type: Opaque
...
 ![img.png](./0000-dns-cluster-aware-delegation-assets/multi-custer-primary-remote-cluster-aware.png)
 
-
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-## CRD Provider
+## Kuadrant Operator
+
+The only change required in the kuadrant operator is to allow the DNSPolices to be created with specifying any provider.
```
maleck13, line 125:
```suggestion
The only change required in the kuadrant operator is to allow the DNSPolices to be created without specifying any provider.
```

maleck13 (approved):
This is in line with what was discussed 👍

PR #134, opened 2025-10-01
Remove references to requestBodyJSON cel function as not supported

eguzki:
### What

Removing references to _requestBodyJSON_ cel function as not supported. More info in https://github.com/Kuadrant/wasm-shim/issues/226

See https://github.com/Kuadrant/kuadrant-operator/pull/1600 for a related change


maleck13 (approved):
@eguzki I am going to go ahead and merge this as I am sure it is valid and should be merged from your input etc on different calls

PR #161, opened 2026-03-23
rfc(grpcroute): add Task 11 for Authorino gRPC well-known attributes

mikenairn:
Authorino needs to extract grpc.service and grpc.method from requests and add them to authorization JSON for use in OPA policies, CEL rules, and other evaluators.

- Add authorino to affected repositories
- Add Task 11 (authorino gRPC WKAs) - can be developed in parallel
- Update from 10 tasks/5 repos to 11 tasks/6 repos
- Update Epic #156 and Tasks 8, 9 dependencies

Authorino follows the same pattern as PR #428 which added the initial well-known attributes support.

Authorino issue created https://github.com/Kuadrant/authorino/issues/584


guicassolato (approved):
Makes sense to me.

PR #163, opened 2026-03-25
rfc: updates to the api design in rfc 0015

guicassolato:
Before:

```yaml
x509:
  source:
    header: "x-forwarded-client-cert" # envoy xfcc format
    -- OR --
    attribute: source.certificate # well-known attribute that resolves to pem-encoded → url-encoded cert
```

After:

```yaml
x509:
  source:
    xfcc: "x-forwarded-client-cert" # envoy xfcc format
    -- OR --
    clientCertHeader: "client-cert" # rfc 9440 format (der-encoded → url-encoded, colon-delimited)
    -- OR --
    expression: request.headers["pem-cert"] # CEL expression that resolves to pem-encoded → url-encoded cert
```

Rationale:
- **Using CEL instead of plain simple WKA:** CEL provides more options to the user, we have all the machinery already, and it's consistent with other features in the API (e.g. `authentication.plain`)
- **Support to RFC 9440:** despite not being widely adopted yet, provides a simpler way for users who need to fallback to Tier 3 compared to encoding the cert info in Envoy's proprietary XFCC format; it's also a new HTTP standard in the making dated from 2023.

### DIFF
--- a/rfcs/0015-x509-client-cert-authpolicy.md
+++ b/rfcs/0015-x509-client-cert-authpolicy.md
@@ -77,7 +77,7 @@ spec:
         # If omitted, defaults to `source.certificate` (CheckRequest attributes)
         source:
           # Option 1: Extract from XFCC header (Envoy format)
-          xfcc: "x-forwarded-client-cert"
+          xfccHeader: "x-forwarded-client-cert"
           # Option 2: Extract from HTTP header (RFC 9440 format)
           # clientCertHeader: "client-cert"
           # Option 3: Extract using CEL expression
@@ -91,7 +91,7 @@ spec:
 ```
 
 **Behavior**:
-- When `source.xfcc` is specified, Authorino extracts the certificate from the named HTTP header using Envoy XFCC format
+- When `source.xfccHeader` is specified, Authorino extracts the certificate from the named HTTP header using Envoy XFCC format
 - When `source.clientCertHeader` is specified, Authorino extracts the certificate from the named HTTP header using RFC 9440 format
 - When `source.expression` is specified, Authorino evaluates the CEL expression to extract the certificate
 - When `source` is omitted, Authorino uses the current behavior of reading from `attributes.source.certificate` (backward compatibility)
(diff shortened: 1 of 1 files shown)


rfcs/0015-x509-client-cert-authpolicy.md
```diff
@@ -320,24 +364,31 @@ type X509AuthenticationSpec struct {
...
 ```
 
 **Default behavior**:
 - If `source` is not specified: backward compatible behavior (read from `attributes.source.certificate`)
-- If `source.header` is specified: extract certificate from the named HTTP header
-- If both `header` and `attribute` are specified: `header` takes precedence
+- If `source.xfcc` is specified: extract certificate from the named HTTP header using Envoy XFCC format
+- If `source.clientCertHeader` is specified: extract certificate from the named HTTP header using RFC 9440 format
+- If `source.expression` is specified: evaluate the CEL expression to extract the certificate
+- If multiple fields are specified: `xfcc` takes precedence, followed by `clientCertHeader`, then `expression`
```
mikenairn: Should these be mutually exclusive instead (validation error if multiple set)?

mikenairn (commented):
Just a question about CEL validations of the source options i.e. should there be some to make them mutually exclusive.

On attribute naming (you mentioned this earlier), only thing i wonder about is the xfcc one not just including "header" in the name (`XFCCHeader string json:"xfccHeader,omitempty` ). 
I don't have a strong opinion on any of this though.

Is the only difference between "XFCC" and "ClientCertHeader" the format of the cert in the header? When using "expression" what format cert is expected?

rfcs/0015-x509-client-cert-authpolicy.md
guicassolato: It's not called out in the design doc, but it's covered in the implementation: https://github.com/Kuadrant/authorino/pull/585/changes#diff-487c9ac14bb7a5eb1def5cf0f2b861c708e8a7836817a3a20c538122e2f0f560R426

guicassolato:
> On attribute naming (you mentioned this earlier), only thing i wonder about is the xfcc one not just including "header" in the name (XFCCHeader string json:"xfccHeader,omitempty ).
I don't have a strong opinion on any of this though.
>
> Is the only difference between "XFCC" and "ClientCertHeader" the format of the cert in the header? When using "expression" what format cert is expected?

This makes sense and it was beginning to haunt me as well. I'll make the change.

mikenairn (approved):

PR #168, opened 2026-04-16
docs(rfc-0017): use map-based gRPC attributes

mikenairn:
Add gRPC well-known attributes to RFC 0017 using a map-based implementation approach.

Changes:
- Add `request.grpc.service` and `request.grpc.method` well-known attributes
- For gRPC requests: `request.grpc` is present as a map with `service` and `method` fields
- For HTTP requests: `request.grpc` key is absent from the request map
- Use standard CEL `has(request.grpc)` macro for presence checking
- Add CEL Presence Checking section with examples and behavior table
- Update Task 10 (wasm-shim) and Task 11 (authorino) descriptions
- Update component responsibilities table
- Add CEL Macros reference for `has()` documentation

Benefits of map-based approach:
- Type safety: Accessing `request.grpc.service` without checking `has(request.grpc)` is a CEL error
- No empty string bugs: Key is absent, not empty
- Standard CEL: Uses built-in `has()` macro; no extension libraries required
- Simplicity: Avoids complexity of alternative approaches
- Consistent: Follows same namespacing pattern as `request.headers`, `request.path`, etc.

## Summary by CodeRabbit

* **Documentation**
  * Added gRPC well-known attributes nested under request.grpc with explicit presence checks (has(request.grpc)) required before access.
  * Added examples, FAQ and predicates showing safe vs error handling for mixed HTTP/gRPC traffic.
  * Added behaviour table clarifying that gRPC metadata is omitted for non-gRPC requests.

## Summary by CodeRabbit

* **Documentation**
  * Updated gRPC route support specification with revised attribute structure and presence-checking semantics.
  * Enhanced guidance for handling non-gRPC requests and malformed paths with corrected implementation examples.

### DIFF
--- a/rfcs/0017-grpcroute-support.md
+++ b/rfcs/0017-grpcroute-support.md
@@ -380,9 +380,9 @@ The following well-known attributes improve UX for `when` clauses and rate limit
   - `request.grpc.method` — extracted from `request.url_path` (e.g., `GetUser` from `/UserService/GetUser`)
 
 **Implementation approach:**
-- WASM shim detects gRPC requests via `content-type: application/grpc` header AND successful parsing of `request.url_path` as `/Service/Method`
+- WASM shim and Authorino (independently) detect gRPC requests via `content-type: application/grpc` header AND successful parsing of `request.url_path` (or path) as `/Service/Method`
 - Path parsing extracts service and method components from `/Service/Method` format
-- `request.grpc` is a map injected into the `request` CEL variable
+- `request.grpc` is a map injected into the `request` CEL variable (wasm-shim) or added to authorization JSON as `input.request.grpc` (Authorino)
 - For gRPC requests: `request.grpc` is present as a map with `service` and `method` string fields
 - For non-gRPC requests OR malformed gRPC paths: `request.grpc` is absent (key does not exist in the request map)
 - Internal predicate generation continues using `request.url_path` to avoid coupling
--- a/rfcs/0017-grpcroute-support.md
+++ b/rfcs/0017-grpcroute-support.md
@@ -390,8 +390,8 @@ The following well-known attributes improve UX for `when` clauses and rate limit
 **Note:** These attributes are **optional** for policy authors. All existing predicates using `request.url_path` continue to work. These well-known attributes provide convenience for policy conditions like:
 ```yaml
 when:
-  - predicate: grpc.hasValue() && grpc.service == 'UserService'
-  - predicate: grpc.hasValue() && grpc.method == 'GetUser'
+  - predicate: grpc.hasValue() && grpc.value().service == 'UserService'
+  - predicate: grpc.hasValue() && grpc.value().method == 'GetUser'
 ```
 
 #### CEL Optional Semantics
@@ -405,25 +405,25 @@ grpc.hasValue()  # Returns true for gRPC requests, false for HTTP
 
 **Safe attribute access:**
 ```yaml
-# Correct: Check hasValue() before accessing fields
+# Correct: Check hasValue() before accessing fields using .value()
 when:
-  - predicate: grpc.hasValue() && grpc.service == 'UserService'
+  - predicate: grpc.hasValue() && grpc.value().service == 'UserService'
 
 # Also valid: Combine with boolean operators
 when:
-  - predicate: grpc.hasValue() && grpc.service == 'UserService' && grpc.method == 'GetUser'
+  - predicate: grpc.hasValue() && grpc.value().service == 'UserService' && grpc.value().method == 'GetUser'
 
 # For negation: Always check hasValue() first
 when:
-  - predicate: grpc.hasValue() && grpc.service != 'AdminService'
+  - predicate: grpc.hasValue() && grpc.value().service != 'AdminService'
 ```
 
 **Benefits of Optional type:**
-- **Type safety**: Accessing `grpc.service` without checking `hasValue()` is a CEL error
+- **Type safety**: Accessing `grpc.value().service` without checking `hasValue()` is a CEL error
 - **No empty string bugs**: Cannot accidentally match HTTP traffic with negation logic
 - **Clear semantics**: `hasValue()` explicitly indicates whether the request is gRPC
-- **Consistent with CEL idioms**: Optional types are a standard CEL pattern for nullable values
-- **Fail-closed security**: Missing `grpc.hasValue()` checks cause CEL evaluation errors at request time, preventing accidental policy bypasses. Unlike empty-string approaches where negation logic (e.g., `grpc.service != 'AdminService'`) could silently match all HTTP traffic, the Optional type ensures policies fail explicitly rather than allowing unintended access.
+- **Consistent with CEL idioms**: Optional types are a standard CEL pattern for nullable values ([CEL Optional Types](https://github.com/google/cel-go/blob/v0.23.2/cel/library.go#L360-L425))
+- **Fail-closed security**: Missing `grpc.hasValue()` checks cause CEL evaluation errors at request time, preventing accidental policy bypasses. Unlike empty-string approaches where negation logic (e.g., `grpc.value().service != 'AdminService'`) could silently match all HTTP traffic, the Optional type ensures policies fail explicitly rather than allowing unintended access.
 
 **Best practice for all policies:** Always include `grpc.hasValue()` checks even when targeting a specific GRPCRoute. While only gRPC traffic flows through GRPCRoute resources, including the check provides defensive coding and prevents errors if policies are later copied to Gateway targets or if routing behavior changes. Consistent use of `hasValue()` makes policy intent explicit and improves maintainability.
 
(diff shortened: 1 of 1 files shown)


rfcs/0017-grpcroute-support.md
```diff
@@ -349,43 +350,56 @@ An empty GRPCRouteMatch (no method, no headers) should return empty predicates,
...
+- `grpc` — CEL Optional binding containing gRPC request metadata (only present for gRPC requests)
+  - `grpc.service` — extracted from `request.url_path` (e.g., `UserService` from `/UserService/GetUser`)
+  - `grpc.method` — extracted from `request.url_path` (e.g., `GetUser` from `/UserService/GetUser`)
 
 **Implementation approach:**
 - WASM shim detects gRPC requests via `content-type: application/grpc` header
 - Path parsing extracts service and method components from `/Service/Method` format
-- Attributes are registered as well-known attributes alongside existing `request.*` attributes
+- `grpc` binding is implemented as a CEL Optional type
+- For gRPC requests: `grpc` contains `optional.of(GrpcMetadata)`
```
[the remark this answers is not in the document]
mikenairn: GrpcMetadata is no longer used in the doc afaik

rfcs/0017-grpcroute-support.md
```diff
@@ -74,8 +74,9 @@ gRPC runs over HTTP/2, not alongside it. From Envoy's perspective, a gRPC reques
 | `request.url_path` | `/api/users/123` | `/package.Service/Method` |
 | `request.headers` | Standard headers | HTTP/2 headers + gRPC metadata |
 | `request.protocol` | `HTTP/1.1` or `HTTP/2` | `HTTP/2` |
-| `grpc.service` | - | `UserService` (from `/UserService/GetUser`) |
-| `grpc.method` | - | `GetUser` (from `/UserService/GetUser`) |
+| `grpc.hasValue()` | `false` | `true` |
```
guicassolato: Is `hasValue()` a standard CEL function?

What's the expected behaviour in the cases where `grpc` is not defined? Will users be required to check `has(grpc) && grpc.hasValue()`?

rfcs/0017-grpcroute-support.md
```diff
@@ -463,10 +478,10 @@ The `kuadrant/testsuite` repository provides the broader E2E test coverage follo
...
+  Implement `grpc` as a CEL Optional binding containing `service` and `method` fields in the WASM shim for improved UX in policy conditions and rate limit counters. For gRPC requests, `grpc` contains `optional.of(metadata)` where metadata includes service and method; for HTTP requests, `grpc` contains `optional.none()`. Update RFC 0002 specification with gRPC well-known attributes documentation.
 
 - [ ] **Task 11: authorino - gRPC Well-Known Attributes** (no blockers - can be developed in parallel)
-  Extract `grpc.service` and `grpc.method` from gRPC requests and add to authorization JSON for use in OPA policies, CEL authorization rules, and other evaluators.
+  Extract `grpc.service` and `grpc.method` from gRPC requests and add to authorization JSON for use in OPA policies, CEL authorization rules, and other evaluators. For gRPC requests, include a `grpc` object with `service` and `method` fields. For non-gRPC requests, the `grpc` field must be absent (not present in the JSON) to align with OPA's `has()` check semantics and avoid serializing unnecessary null values.
```
guicassolato: I believe `has()` is a CEL function. It's not a Rego's [built-in](https://www.openpolicyagent.org/docs/policy-reference/builtins) AFAIK, although [`is_null`](https://www.openpolicyagent.org/docs/policy-reference/builtins/types) exists in some environments not tested with Authorino.

rfcs/0017-grpcroute-support.md
mikenairn: > Is `hasValue()` a standard CEL function?

Yes, `hasValue()` is a standard CEL function from the `cel.OptionalTypes()` library ([docs](https://github.com/google/cel-go/blob/v0.23.2/cel/library.go#L213)).

> What's the expected behaviour in the cases where `grpc` is not defined? Will users be required to check `has(grpc) && grpc.hasValue()`?

When a variable is declared and bound in the activation context, it's always accessible - no `has(grpc)` check needed.

There was an issue with the example syntax is was using though, CEL requires `.value()` to access Optional contents:                                                                                                                                                       
                                                                  
  ```yaml                                                                                                                                                                                                                                                                                                                                    
  grpc.hasValue() && grpc.value().service == 'UserService'
```

File: rfcs/0017-grpcroute-support.md (hunk 3 above)
Comment (mikenairn) [reply rfcs/0017-grpcroute-support.md]: Yes `has()` is a CEL macro, not a Rego built-in. Updated the wording as suggested.

File: rfcs/0017-grpcroute-support.md (hunk 2 above)
Comment (mikenairn) [reply rfcs/0017-grpcroute-support.md]: Actually, I'm going to double check all this.

File: rfcs/0017-grpcroute-support.md [hunk 4]
```diff
@@ -74,8 +74,9 @@ gRPC runs over HTTP/2, not alongside it. From Envoy's perspective, a gRPC reques
 | `request.url_path` | `/api/users/123` | `/package.Service/Method` |
 | `request.headers` | Standard headers | HTTP/2 headers + gRPC metadata |
 | `request.protocol` | `HTTP/1.1` or `HTTP/2` | `HTTP/2` |
-| `grpc.service` | - | `UserService` (from `/UserService/GetUser`) |
-| `grpc.method` | - | `GetUser` (from `/UserService/GetUser`) |
+| `grpc.hasValue()` | `false` | `true` |
+| `grpc.service` | Error (optional.none()) | `UserService` (from `/UserService/GetUser`) |
+| `grpc.method` | Error (optional.none()) | `GetUser` (from `/UserService/GetUser`) |
```
averevki:
```suggestion
| `grpc.service` | Error (`optional.none()`) | `UserService` (from `/UserService/GetUser`) |
| `grpc.method` | Error (`optional.none()`) | `GetUser` (from `/UserService/GetUser`) |
```
does this break the table formatting?

rfcs/0017-grpcroute-support.md
mikenairn: `hasValue()` is actually an "Optional" CEL method and not a standard function from what i can tell. Even though both the GO and Rust impls support it,  I'm going to avoid using it here, instead resorting to a nil map value and the use of `has(request.grpc)`

rfcs/0017-grpcroute-support.md
```diff
@@ -374,43 +375,66 @@ An empty GRPCRouteMatch (no method, no headers) should return empty predicates,
...
+- **Clear semantics**: `has(request.grpc)` explicitly indicates whether the request is gRPC
+- **Standard CEL**: Uses the built-in `has()` macro; no CEL Optional packages required. While CEL Optional types (`hasValue()`, `value()`) are supported in the Rust CEL library, the map-based approach avoids the complexity of Optional type handling.
+- **Fail-closed security**: Missing `has(request.grpc)` checks cause CEL evaluation errors at request time, preventing accidental policy bypasses. The absent key ensures policies fail explicitly rather than allowing unintended access.
+- **Consistent with other attributes**: Follows the same namespacing pattern as `request.headers`, `request.path`, etc.
+
+**Best practice for all policies:** Always include `has(request.grpc)` checks even when targeting a specific GRPCRoute. While only gRPC traffic flows through GRPCRoute resources, including the check provides defensive coding and prevents errors if policies are later copied to Gateway targets or if routing behavior changes. Consistent use of `has()` makes policy intent explicit and improves maintainability.
```
mikenairn: PR title and description have been updated

rfcs/0017-grpcroute-support.md
```diff
@@ -374,43 +375,66 @@ An empty GRPCRouteMatch (no method, no headers) should return empty predicates,
...
 
-- `grpc.service` — extracted from `request.url_path` (e.g., `UserService` from `/UserService/GetUser`)
-- `grpc.method` — extracted from `request.url_path` (e.g., `GetUser` from `/UserService/GetUser`)
+- `request.grpc` — nested map containing gRPC request metadata, accessible via standard CEL `has()` macro
+  - `request.grpc.service` — extracted from `request.url_path` (e.g., `UserService` from `/UserService/GetUser`)
+  - `request.grpc.method` — extracted from `request.url_path` (e.g., `GetUser` from `/UserService/GetUser`)
 
 **Implementation approach:**
-- WASM shim detects gRPC requests via `content-type: application/grpc` header
+- WASM shim detects gRPC requests via `content-type: application/grpc` header AND successful parsing of `request.url_path` as `/Service/Method`
```
guicassolato: Should we include `&& request.protocol == "HTTP/2"` too? I'm not sure about the situation of GRPCRoute for gRPC-Web.

guicassolato: We should probably include Authorino here:

OPTION 1:

```suggestion
- WASM shim and Authorino (independently) detect gRPC requests via `content-type: application/grpc` header AND successful parsing of `request.url_path` as `/Service/Method`
```

OPTION 2:

```suggestion
- WASM shim detects gRPC requests via `content-type: application/grpc` header AND successful parsing of `request.url_path` as `/Service/Method`
- WASM shim hints Authorino about the GRPC request, by specifying additional context metadata that Authorino will then use to enable support for `request.grpc`
```

mikenairn: Apparently gRPC-web can be HTTP/1.1 or HTTP/2 http https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md. So we could add it, but there is that. 
We could also ensure the http method is POST as well if you are trying to narrow down the scope of matches, but tbh it would be malformed anyway if it is actually gRPC so ....

mikenairn: Updated as described in Option 1

guicassolato (approved):
Thanks, @mikenairn! This is great.

About gRPC-Web, at least Envoy Gateway supports it with the GRPCRoute apparently: https://gateway.envoyproxy.io/v1.5/tasks/traffic/grpc-routing/. I didn't check where Istio stands, but one is more than enough not to filter based on the protocol for now IMO.

rfcs/0017-grpcroute-support.md
mikenairn: Will not check for protocol for now see https://github.com/Kuadrant/architecture/pull/168#pullrequestreview-4187180965

rfcs/0017-grpcroute-support.md
mikenairn: table looks fine, think this is a stale review comment

rfcs/0017-grpcroute-support.md
alexsnaps: Not that this matters really anymore, but:
>`hasValue()` is actually an "Optional" CEL method and not a standard function from what i can tell.

It is a standard, yet optional _syntax_ and [related functions](https://github.com/google/cel-go/blob/c9d70486a872fbfaf9c4cb383a005de56e499368/cel/env_test.go#L284-L304), but they are [part of the grammar](https://github.com/google/cel-go/blob/master/parser/gen/CEL.g4#L55). They can be disabled/enabled as per how a user would want to use these.

>CEL requires .value() to access Optional contents

No, you can just deref the value transparently, _should a value be there_, eg [here](https://playcel.undistro.io/?content=H4sIAAAAAAAAA6tWSk7NUbJSyi8oyczPS8zRy0%2FTqI5RSsvPj1GyUjCs1dQDMpV0lFISSxI98wpKS4BqY%2FKAArn5KalANkh3LQDyw5uZRQAAAA%3D%3D):

`optional.none().foo` would return `null`, so would `optional.none().foo.bar` and so forth

PR #171, opened 2026-04-21
RFC 0019: Graduate components from 0.x to v1.0

thomasmaas:
## Summary

- Graduate all 0.x Kuadrant components to v1.0.0 in a single release cycle
- Signals API stability and makes semver meaningful for automation
- No code or API changes — just a version bump
- Components affected: authorino, authorino-operator, limitador-operator, dns-operator, wasm-shim, console-plugin, policy-machinery

## Motivation

kuadrant-operator is at v1.4 and limitador at v2.3, but 7 of 9 suite components are still on 0.x despite having stable APIs and production usage. Under semver, 0.x means "anything may change" — which contradicts reality and prevents tooling from trusting version signals.

## Test plan

- [ ] Review whether each component's API surface is ready for a v1.0 stability commitment
- [ ] Discuss timing — should all graduate in the same kuadrant release cycle?
- [ ] Decide on wasm-shim and policy-machinery (see unresolved questions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


thomasmaas:
Some newer components might not be ready for this move. dns-operator and console-plugin come to mind.

rfcs/0019-semver-v1-graduation.md
```diff
@@ -0,0 +1,132 @@
+# Graduate Components to v1.0
+
+- Feature Name: `semver-v1-graduation`
+- Start Date: 2026-04-21
+- RFC PR: TBD
```
Boomatang, line 5: This can be filled in now.

rfcs/0019-semver-v1-graduation.md
```diff
@@ -0,0 +1,132 @@
...
+
+| Component | Current | Graduates to | Notes |
+|-----------|:-------:|:------------:|-------|
+| authorino | v0.25.0 | v1.0.0 | |
+| authorino-operator | v0.24.0 | v1.0.0 | |
+| limitador-operator | v0.17.1 | v1.0.0 | |
+| dns-operator | v0.16.1 | v1.0.0 | |
+| wasm-shim | v0.12.3 | v1.0.0 | |
+| console-plugin | v0.3.6 | v1.0.0 | |
+| policy-machinery | v0.8.0 | v1.0.0 | Go module path stays `github.com/kuadrant/policy-machinery` (no `/v2` needed since graduating from 0.x to 1.x) |
```
Boomatang, line 65: Of every thing, I am not sure about the policy-machinery. That is the one thing that I would call the least stable from the fact we haven't had to do a lot of work there, but that does not mean there is not work needed.

Boomatang (commented):
I would be on board with this change. The policy-machinery is the only questionable I would have.

PR #178, opened 2026-06-05
RFC 0020: Two-Phase Release Workflow

Boomatang:
# Summary

This RFC introduces a two-phase release workflow model for all Kuadrant component repositories, evolving the process described in RFC 0008.

The v1.5.0 release exposed systemic failures in the single-workflow release model: false success reporting, hidden workflow chains, silent downstream failures, policy violations from direct commits to protected branches, and missing dependency gating. These all stem from conflating code changes with artifact publishing in a single workflow.

# What this RFC proposes

Every release is split into two distinct GitHub Actions workflows:

- A **pre-release** workflow that makes version-related code changes and opens a pull request to the release branch. No tags or versioned artifacts are created.
- A **release** workflow that runs smoke tests, tags, builds all artifacts, and creates the GitHub Release as its final step. No code changes are made.

A human review gate (PR approval) sits between the two phases, satisfying Red Hat's Segregation of Duties policy and giving CI an opportunity to validate the release branch before any artifacts are published.

A `release.yaml` file is added to every component repository as the machine-readable source of truth for version and dependency information. A **version gate** CI check uses this file to prevent operators from releasing before their operand dependencies have published GitHub Releases.

# Key guarantees

- No phantom releases: if any step fails, no GitHub Release exists
- Tests run before any artifacts are built or published
- All code changes go through PR review before release proceeds
- Dependency ordering is enforced automatically via the version gate
- Each component's release is self-contained with no direct cross-repo workflow triggers
- Branch protection compliance: the release workflow only creates a tag, never commits to protected branches

# Relationship to existing RFCs

- Evolves the workflow mechanics of RFC 0008 (Kuadrant Release Process) without replacing its broader concerns (versioning, cadence, QA handover, communication)
- Builds on the `release-X.Y` branch naming convention defined in RFC 0018 (Standardize Release Branch Naming)
- Validated through a reference implementation at [kuadrant-labs/workflow_layout](https://github.com/kuadrant-labs/workflow_layout)

## Summary by CodeRabbit

* **Documentation**
  * Defined a standardised two-phase release workflow for component repositories.
  * Added a manually triggered pre-release phase that validates versions, updates release information, and opens a release pull request without publishing artefacts.
  * Added a release phase that runs tests before creating tags, publishing artefacts in dependency order, and creating the GitHub Release.
  * Documented dependency checks, version semantics, branch naming, and compliance requirements.

### DIFF
--- a/rfcs/0020-two-phase-release-workflow.md
+++ b/rfcs/0020-two-phase-release-workflow.md
@@ -3,7 +3,7 @@
 - Feature Name: `two_phase_release_workflow`
 - Start Date: 2026-06-04
 - RFC PR: [Kuadrant/architecture#178](https://github.com/Kuadrant/architecture/pull/178)
-- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+- Issue tracking:
 
 # Summary
 [summary]: #summary
@@ -131,7 +131,7 @@ Operators declare their operand dependencies, so the version gate can verify ava
 
 Release branches follow the convention `release-X.Y` where `X.Y` is the minor version, without a `v` prefix.
 For example, version `1.5.0` releases from branch `release-1.5`, and a subsequent patch `1.5.1` releases from the same branch.
-This convention is defined in detail in [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-rlease-branch-naming.md).
+This convention is defined in detail in [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-release-branch-naming.md).
 
 Using minor-level branches means patch releases do not require new branches - they reuse the existing release branch with an updated version in `release.yaml`.
 This is critical for automation: a consistent branch naming scheme allows tooling to discover and operate on release branches without per-component configuration.
@@ -203,8 +203,9 @@ dependencies:
 
 - `<component-name>` (required): The name of the component. Must match the repository name or the component's canonical identifier.
 - `version` (required): A semantic version string (`X.Y.Z`), or `0.0.0` on the `main` branch to indicate active development.
-- `dependencies` (optional): A map of dependency names to their required versions. Each dependency name must correspond to a repository in the Kuadrant GitHub organization.
-The version `0.0.0` indicates "targets latest" and is not validated by the version gate.
+- `dependencies` (optional): A map of dependency names to their required versions. 
+Each dependency name must correspond to a repository in the Kuadrant GitHub organization.
+The version `0.0.0` indicates "targets latest" and is rejected by the version gate, see [Version gate specification](#Version-gate-specification).
 
 **Invariants:**
 
--- a/rfcs/0020-two-phase-release-workflow.md
+++ b/rfcs/0020-two-phase-release-workflow.md
@@ -131,7 +131,7 @@ Operators declare their operand dependencies, so the version gate can verify ava
 
 Release branches follow the convention `release-X.Y` where `X.Y` is the minor version, without a `v` prefix.
 For example, version `1.5.0` releases from branch `release-1.5`, and a subsequent patch `1.5.1` releases from the same branch.
-This convention is defined in detail in [RFC 0015](https://github.com/Kuadrant/architecture/blob/main/rfcs/0015-standardize-release-branch-naming.md).
+This convention is defined in detail in [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-rlease-branch-naming.md).
 
 Using minor-level branches means patch releases do not require new branches - they reuse the existing release branch with an updated version in `release.yaml`.
 This is critical for automation: a consistent branch naming scheme allows tooling to discover and operate on release branches without per-component configuration.
@@ -467,7 +467,7 @@ This is a separate initiative and does not affect the model defined here.
 
 - **Release branch naming standardization.**
 Full standardization of the `release-X.Y` branch naming convention across all repositories, including migration of existing branches that use different conventions, is tracked as a companion effort.
-See [RFC 0015](https://github.com/Kuadrant/architecture/blob/main/rfcs/0015-standardize-release-branch-naming.md) for details.
+See [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-release-branch-naming.md) for details.
 
 - **Automated version gate expansion.**
 The version gate currently checks for GitHub Release existence.
--- a/rfcs/0020-two-phase-release-workflow.md
+++ b/rfcs/0020-two-phase-release-workflow.md
@@ -212,6 +212,8 @@ The version `0.0.0` indicates "targets latest" and is not validated by the versi
 - On release branches, the version must not be `0.0.0`.
 - On release branches, dependency versions must not be `0.0.0` - they must be concrete released versions.
 The version gate will verify that a GitHub Release exists for each dependency.
+- Where the component's build system maintains its own version variables (e.g., Makefile variables), these must reflect the values in `release.yaml` on release branches.
+The pre-release workflow is responsible for ensuring this consistency.
 
 ### Pre-release workflow specification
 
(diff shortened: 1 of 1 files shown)


rfcs/0020-two-phase-release-workflow.md
```diff
@@ -0,0 +1,477 @@
...
+## Workflow for a release engineer
+
+A typical minor release follows these steps:
+
+1. Trigger the pre-release workflow with the target version (e.g., `1.5.0`).
+2. The workflow creates the release branch `release-1.5` (if needed), updates `release.yaml`, runs component-specific preparation steps, and opens a PR.
+3. Review the PR. CI runs tests and the version gate check. 
+If the component is an operator, the version gate confirms operand releases exist.
+4. Merge the PR.
+5. Trigger the release workflow with the release branch (`release-1.5`).
```
guicassolato, line 168: Do you see value in adding a grace period between pre-release workflow and release workflow?

Boomatang: It depends on what you define as a grace period. The PR will require approval, and would need to a normal PR review. The reviewer would not be the same person that created the release PR. 

Or are you talking about a grace period between the PR generated form the pre-release workflow being merged and the release workflow being triggered. There could be value there. I have had some conversation on how it could be possible for QE to do a round of test before release are generated, but there has being no concrete view of what that process would be. 

So as of today I don't see a reason to force a grace period, but open to amends later if there is value in having a grace period. I don't see any thing blocking the adding of a grace period at a later date, and I don't see the need of adding one today.

guicassolato: > Or are you talking about a grace period between the PR generated form the pre-release workflow being merged and the release workflow being triggered.

Yes

> as of today I don't see a reason to force a grace period

OK. It doesn't have to be a hard requirement. It could start as an informal practice at first and then, if we need to formalise it, we can do it.

rfcs/0020-two-phase-release-workflow.md
```diff
@@ -0,0 +1,477 @@
...
+This is actively being designed but is independent of this RFC.
+
+- **Centralized GitHub Actions.**
+The duplicated workflow structure across repositories could be consolidated into shared composite actions and reusable workflows in a central repository.
+This would reduce the maintenance burden of updating workflow logic across all components.
+This is a separate initiative and does not affect the model defined here.
+
+- **Release branch naming standardization.**
+Full standardization of the `release-X.Y` branch naming convention across all repositories, including migration of existing branches that use different conventions, is tracked as a companion effort.
+See [RFC 0015](https://github.com/Kuadrant/architecture/blob/main/rfcs/0015-standardize-release-branch-naming.md) for details.
```
averevki: This should be the RFC 0018 instead. Another wrong reference was also pointed out by coderabbit above

rfcs/0020-two-phase-release-workflow.md
```diff
@@ -0,0 +1,477 @@
...
+  <dependency-name>: "<semver>"
+  ...
+```
+
+**Fields:**
+
+- `<component-name>` (required): The name of the component. Must match the repository name or the component's canonical identifier.
+- `version` (required): A semantic version string (`X.Y.Z`), or `0.0.0` on the `main` branch to indicate active development.
+- `dependencies` (optional): A map of dependency names to their required versions. Each dependency name must correspond to a repository in the Kuadrant GitHub organization.
+The version `0.0.0` indicates "targets latest" and is not validated by the version gate.
```
mikenairn: This file will duplicate values that are already defined as Makefile variables. If another file is required to store this data, they should at least be compatible. Make targets triggered on these branches must use the same values.

mikenairn: Any reason not to use the pattern already established in some of the repos: 
* https://github.com/Kuadrant/limitador-operator/blob/v0.18.2/make/release.mk 
* https://github.com/Kuadrant/dns-operator/blob/v0.17.0/make/release.mk

Boomatang: The existing release.mk pattern works well for what it does, providing Make variables for build targets in side Go projects. This RFC isn't trying to replace that.

The problem release.yaml solves is different: cross-repo dependency gating. The version gate needs to check whether an operator's operand dependencies have  published GitHub Releases before the operator can release. That requires a machine-readable file with a consistent schema across every repo — regardless of  whether the project uses Make, Cargo, npm, or something else.

` release.mk` is Make-specific and its structure varies between repos that use it. It can't serve as a common interface for external tooling that needs to walk  the dependency tree across all components.

You're right that the values must stay consistent. A project's Makefile targets should consume the version from release.yaml (or the pre-release workflow should update both).

mikenairn: > You're right that the values must stay consistent. A project's Makefile targets should consume the version from release.yaml (or the pre-release workflow should update both).

Should this be acknowledged in the doc? There should only really be one source of truth for these values IMO. The make targets on the branches should also continue to work as expected (using correct values for the particular branch).

The kaudrant-operator for instance that already has this [file](https://github.com/Kuadrant/kuadrant-operator/blob/v1.5.0/release.yaml#L2) doesn't appear to have working [make targets](https://github.com/Kuadrant/kuadrant-operator/blob/v1.5.0/Makefile#L22). Unless I'm missing something, seems impossible to build a local developer version from a release branch since none of the [dependency versions](https://github.com/Kuadrant/kuadrant-operator/blob/v1.5.0/release.yaml#L8-L13) are updated in Make.

Boomatang: I don't know if I am fully following you here. If you checkout the v1.5.0 release tag and modify the version in the `release.yaml` and then call `make prepare-release` the versions are updated. Or are you talking about some other bug, if so can you share a link to the issue.

mikenairn: The issue is developer experience, if I checkout v1.5.0 and run make local-setup, the Makefile defaults all dependency versions to latest/main because it never reads release.yaml. I silently get the wrong dependencies with no error. A developer shouldn't need to know about the release pipeline to build from a release branch.

dns-operator and limitador-operator solved this with make/release.mk, a generated file committed to release branches that overrides Make defaults. Every make target works correctly on every branch with zero extra steps. If release.yaml is being standardised, the RFC should specify that the pre-release workflow generates build-system-native overrides from it (e.g. make/release.mk for Go projects), so there's one source of truth (release.yaml) with a derived artefact that keeps Make working.

Boomatang: Yes, I think we are both on the same page. The release.yaml will be the standardised interface for external tooling to interact with a repo. I didn't want to prescribe in the RFC that Go projects need to do X, x in this case generating a make/release.mk file. That was a project specific implementation detail that I felt was out side the scope of the RFC, but as the concern has being raised I don't mind adding a comment in.

Boomatang: I have added https://github.com/Kuadrant/architecture/pull/178/changes/9cb12fb34f472128ac340b8f6dcd084ba861b9f7 which I am hoping will address your concerns.

rfcs/0020-two-phase-release-workflow.md
```diff
@@ -0,0 +1,481 @@
...
```
averevki: Should this be something like `rejected if 0.0.0 is set for any operator or operand versions` instead of `not validated`?

rfcs/0020-two-phase-release-workflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+The `dependencies` section is optional. 
+Leaf components (operands with no Kuadrant dependencies) omit it.
+Operators declare their operand dependencies, so the version gate can verify availability before the release proceeds.
+
+## Release branch naming
+
+Release branches follow the convention `release-X.Y` where `X.Y` is the minor version, without a `v` prefix.
+For example, version `1.5.0` releases from branch `release-1.5`, and a subsequent patch `1.5.1` releases from the same branch.
+This convention is defined in detail in [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-rlease-branch-naming.md).
```
averevki:
```suggestion
This convention is defined in detail in [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-release-branch-naming.md).
```

rfcs/0020-two-phase-release-workflow.md
```diff
@@ -0,0 +1,481 @@
+# Two-Phase Release Workflow
+
+- Feature Name: `two_phase_release_workflow`
+- Start Date: 2026-06-04
+- RFC PR: [Kuadrant/architecture#178](https://github.com/Kuadrant/architecture/pull/178)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
```
averevki: Is there any issue we can put here instead of this placeholder?

Boomatang: There is no Issue tracker so I have removed the link

rfcs/0020-two-phase-release-workflow.md
Boomatang: Did a small rewording.

averevki (approved):

PR #179, opened 2026-06-08
Olmv1 support

### LINKED ISSUE #164: RFC: OLMv1 support - replace OLM dependency model with operator-managed ClusterExtensions
## Summary

OLMv1 removes automatic dependency installation. The Kuadrant operator currently relies on OLMv0 dependency model to automatically install Authorino, Limitador, and DNS operators when a user installs Kuadrant via a single Subscription.

With OLMv1, each operator must be installed as a separate ClusterExtension by the cluster admin, requiring ~20+ resources to install the full Kuadrant stack manually.

maleck13:
RFC for replacing OLMv0 dependency model with an umbrella operator.

  Key decisions:
  - Helm charts become the single manifest source for both install paths (direct Helm on k8s,
  umbrella operator on OpenShift)
  - Umbrella operator renders charts client-only (no Helm release tracking), same pattern as
  cluster-olm-operator
  - Operand creation (Authorino, Limitador) moves out of kuadrant-operator to the Helm
  chart/umbrella operator
  - Kuadrant CR owned by new umbrella operator
  - kuadrant operator becomes pure policy controller
  - Upgrade ordering: CRDs/RBAC first, dependency operators second, kuadrant-operator last
  - Kuadrant operator becomes purely policy controller

  Closes #164

## Summary by CodeRabbit

* **Documentation**
  * Added RFC proposing OLMv1 deployment consolidation for Kuadrant.
  * Outlines transition to Helm-based installation as the primary deployment mechanism.
  * Documents planned internal API changes and migration strategy for existing deployments.


rfcs/0018-olmv1-support.md
```diff
@@ -0,0 +1,181 @@
...
+
+This pattern extends to any future operator that builds on Kuadrant's policy CRDs — the umbrella operator provides the install ordering guarantee that OLMv0 dependency resolution previously handled.
+
+### OLM Manifest Removal
+
+With the umbrella operator owning deployment of all components via Helm charts, individual operators no longer need their own OLM packaging artifacts (CSV, bundle manifests, catalog entries, `bundle.Dockerfile`). These can optionally be removed from each operator repo, leaving only the Helm chart as the manifest source.
+
+This is **optional per operator**. Operators such as authorino-operator and limitador-operator have standalone use cases outside of a Kuadrant installation and may choose to retain their own OLM bundles for independent installation via OLMv1. Operators that are only used as part of a Kuadrant deployment (e.g., mcp-gateway-operator, kuadrant-operator) are stronger candidates for removal.
+
```
mikenairn: :tada: This IMO is significant and i would personally rather see this is a *Must* to remove all OLM complexity from individual component operator repos.

rfcs/0018-olmv1-support.md
```diff
@@ -0,0 +1,181 @@
...
+2. Create the umbrella operator's ClusterExtension (with namespace, ServiceAccount, RBAC). This is the only ClusterExtension the user creates.
+3. OLMv1 deploys the umbrella operator.
+4. The umbrella operator renders the Helm charts and deploys all components: CRDs, RBAC, operator Deployments (including extension operators such as mcp-gateway-operator), and operand instances.
+5. Create the Kuadrant CR (to configure) — kuadrant-operator begins reconciling policies.
+
+Uninstall: delete the Kuadrant CR , then delete the ClusterExtension.
+
+### Upgrade
+
+#### Initial transition (from OLMv0 to umbrella operator)
```
Boomatang: I am wondering what the upgrade path looks like during the initial transition state. My understand here is you are expected this to be a phased approach where the kuadrant-operator installations via the umbrella operator is made stable first before doing the operands. 

Does this mean for a time it is possible to have some operands to be installed via OLMv0, if that is case what is the approach for taken ownership of the resources? Who is responsible for cleaning up OLMv0 resources?

rfcs/0018-olmv1-support.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Requirements
+
+1. **Single-action install** — users create one ClusterExtension, not four.
+2. **Full lifecycle management** — the umbrella operator manages versions, upgrades, and removal of all components.
+3. **OLMv1 compatible** — works with the ClusterExtensions API on both OpenShift and vanilla Kubernetes.
+4. **Separation of concerns** — the umbrella operator handles deployment; kuadrant-operator handles policy reconciliation and has no OLMv1 awareness.
+5. **Helm as single manifest source** — both installation paths (direct Helm (upstream plain kubernetes) and umbrella operator (OpenShift)) consume the same charts.
+6. **Umbrella operator owns operands** — installs each dependency operator and creates their operands (Authorino, Limitador instances). The kuadrant-operator no longer performs these tasks.
+7. **Extension operators managed centrally** — operators that depend on kuadrant-operator (e.g., mcp-gateway-operator) are deployed by the umbrella operator after kuadrant-operator, removing their need for OLM dependency declarations.
```
Boomatang: > and creates their operands (Authorino, Limitador instances)

I am assuming this is via their CRDs. The Limitador CR is one that I have question about in this instances. Who has ownership of the CR? The configuration for the limits are also in the limitador CR, so we will have both umbrella operator and kuadrant-operator trying to maintain the configuration.

Boomatang (commented):
I like how this will simplify the operators, and put helm to the for-forth. I have a concern that the umbrella operator will be an operator of operators which will have to deal with upgrade paths. In general I don't think we focus enough on the upgrade paths. There could easily be assumptions here that we are making without knowing.

rfcs/0018-olmv1-support.md
maleck13: Yes that mix of concerns is a frustration but is probably beyond this work to change (As in have separated limits). My initial thought here is that the Umbrella operator owns the resource but doesn't own the limits (if you get me). So wether the resource exits at all is up to the umbrella operator. What is in the limits is up to the kuadrant operator based on the policies. It is then an open question as to whether we want to eventually move these limit definitions out of the shared resource. WDYT

rfcs/0018-olmv1-support.md
maleck13: Thanks @Boomatang I am working through the specifics of upgrade. It may be that what we want is to introduce the umbrella operator via olmv0 also . I will add more details here around upgrade / migration and let you know once it is there

rfcs/0019-olmv1-operator-consolidation.md
Boomatang, line 1: Has there being scope creep in this? It now seems to be suggestion that we are dropping support for authorino-operator and the limitador-operator, with a rewrite of their logic into the kuadrant-operator. With the authorino-operator in particular, is there much of community usage of that operator? I assume we will be having to put a deprecation notice on the operators. Or is the idea to have both the logic in the kuadrant-operator, and have the standalone operators.

maleck13: the scope creep here is dropping use of the operators in Kuadrant rather than deleting them entirely.

mikenairn (approved):
Good starting point. Some of the details will change over the next while as we progress some of the exploratory tasks but we can update as required.

PR #189, opened 2026-07-17
RFC 0019: Kuadrant Operator Consolidation

mikenairn:
## Summary

Updates RFC 0019 to reflect the operator consolidation approach. Validated by POC branches on `kuadrant-operator` (`olmv1-umbrella-poc-phase1`, `olmv1-umbrella-poc-static-manifests`) and tested on OpenShift CRC with both OLMv0 and OLMv1.

### Approach

Child operator Helm charts are embedded in the kuadrant-operator container image and rendered at runtime on startup. All child operator resources (CRDs, ClusterRoles, Deployments, SAs, RBAC) are applied by the operator before the controller manager starts. Nothing from child operators is included in the OLM bundle, eliminating GVK migration conflicts entirely.

This follows the same pattern used by Red Hat OpenShift AI ([opendatahub-operator](https://github.com/opendatahub-io/opendatahub-operator)) and OpenShift's [cluster-olm-operator](https://github.com/openshift/cluster-olm-operator), and is the approach recommended by the OLM team for operators that previously relied on dependency declarations.

### Key decisions

- Child operators are control plane singletons deployed in the operator namespace at startup, not triggered by Kuadrant CR
- Charts consumed as-is from upstream repos, no rendering or splitting at sync time
- OLM bundle contains only kuadrant-operator's own CRDs and resources
- OLM migration handled programmatically (operator cleans up orphaned child Subscriptions/CSVs)
- OLMv1 compatibility verified via ClusterExtension on CRC
- RBAC: scoped CRD permissions (unscoped `create`, scoped `get`/`update`/`patch`), `escalate` for ClusterRoles, `bind` for ClusterRoleBindings
- Kuadrant CR only triggers data plane (wrapper CRs, policies)

### Alternatives considered and rejected

- Installer-managed CRDs (split chart approach): OLM GVK conflicts, divergent deployment model
- Consolidated CSV (large CSV): OLM GVK conflicts, no path to conditional deployment
- Operator-managed Subscriptions: OLMv0-only API, two separate code paths for OLM/Helm
- New umbrella operator: unnecessary new codebase
- Manual dependency installation: UX regression

### POC Branches

- `kuadrant-operator`: [`olmv1-umbrella-poc-phase1`](https://github.com/Kuadrant/kuadrant-operator/tree/olmv1-umbrella-poc-phase1) (runtime rendering)
- `kuadrant-operator`: [`olmv1-umbrella-poc-static-manifests`](https://github.com/Kuadrant/kuadrant-operator/tree/olmv1-umbrella-poc-static-manifests) (consolidated CSV alternative)
- `dns-operator`: [`olmv1-umbrella-poc-phase1`](https://github.com/Kuadrant/dns-operator/tree/olmv1-umbrella-poc-phase1)

### DIFF
--- a/rfcs/0019-olmv1-operator-consolidation.md
+++ b/rfcs/0019-olmv1-operator-consolidation.md
@@ -138,7 +138,11 @@ The operator does not create or modify cluster-scoped resources (CRDs, ClusterRo
 
 When a user creates a Kuadrant CR, the kuadrant-operator renders each component controller's Helm chart from `/charts/<name>/` in the container image using the Helm Go SDK (`ClientOnly=true`, `DryRun=true`). The rendered namespaced resources are applied via Server-Side Apply with the kuadrant-operator as the field manager.
 
+Component controllers are deployed into the Kuadrant CR's namespace, not the kuadrant-operator's namespace. This is a change from the current model where all child operators run in `kuadrant-system` regardless of the Kuadrant CR's location. This approach moves towards supporting future multi-tenanted control plane scenarios.
+
 All component controller Deployments are owned by the Kuadrant CR via ownerReference. This includes dns-operator, which previously ran independently and did not require a Kuadrant CR.
+
+Observability (metrics collection, ServiceMonitors, log aggregation) must account for component controllers running in the Kuadrant CR's namespace rather than a fixed operator namespace.
 
 ## RBAC
 
--- a/rfcs/0019-olmv1-operator-consolidation.md
+++ b/rfcs/0019-olmv1-operator-consolidation.md
@@ -26,6 +26,7 @@ Consolidate the Kuadrant ecosystem into a single operator. The kuadrant-operator
 2. **Removing child operator controllers**: authorino-operator and limitador-operator continue to run as Deployments. Deferred to a follow-on RFC
 3. **Migrating to OLMv1**: all changes run on OLMv0. OLMv1 compatibility is a readiness outcome, not a migration step
 4. **Refactoring component Helm charts**: upstream charts are consumed as-is. Improved templating is future work
+5. **Conditional component deployment**: all components are always deployed. Optional component installation is future work
 
 # Motivation
 [motivation]: #motivation
@@ -36,9 +37,9 @@ OLMv1 [explicitly removes automatic dependency installation](https://github.com/
 
 > OLM v1 will not automatically install a missing dependency when a user requests an operator. The primary reasoning is that OLM v1 will err on the side of predictability and cluster-administrator awareness.
 
-Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint` will be rejected under OLMv1. Kuadrant's current model relies entirely on these declarations.
+Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint` will be rejected under OLMv1. Kuadrant's current model relies entirely on these declarations. This has been verified on OpenShift CRC: attempting to install the current kuadrant-operator via OLMv1 `ClusterExtension` fails with `bundle has a dependency declared via property "olm.package.required" which is currently not supported`.
 
-The exact process for transitioning a cluster (fully) from OLMv0 to OLMv1 is not yet documented — this is an open question that affects all operators, not just Kuadrant. Currently on OpenShift OLMv1 and OLMv0 run alongside each other. However, regardless of when or how that transition happens, the dependency resolution mechanism Kuadrant relies on will not exist in the future. We should eliminate our dependency declarations now, while still on OLMv0, so that Kuadrant is already compatible when the transition arrives.
+The exact process for transitioning a cluster (fully) from OLMv0 to OLMv1 is not yet documented. Currently on OpenShift OLMv1 and OLMv0 run alongside each other. However, regardless of when or how that transition happens, the dependency resolution mechanism Kuadrant relies on will not exist in the future. We should eliminate our dependency declarations now, while still on OLMv0, so that Kuadrant is already compatible when the transition arrives.
 
 ## Per-component OLM packaging adds overhead within a Kuadrant deployment
 
@@ -54,6 +55,10 @@ Kuadrant's deployment targets are not limited to OpenShift. In the future, we wa
 
 Each independent OLM operator requires its own bundle image, catalog entry, release pipeline, and version coordination. Consolidating into a single OLM package eliminates per-component release overhead while keeping the same controllers and workloads running on the cluster. No CRDs, Deployments, or other resources are removed or deprecated as part of this work.
 
+## Established pattern for OLMv1 readiness
+
+The meta-operator pattern (a parent operator deploying child components at runtime) is the approach highlighted by the OLM team for operators that previously relied on `olm.package.required` dependencies. It is already used in production by Red Hat OpenShift AI ([opendatahub-operator/rhods-operator](https://github.com/opendatahub-io/opendatahub-operator)) for 18+ components, and by the [cluster-olm-operator](https://github.com/openshift/cluster-olm-operator) which manages OLMv1's own sub-components using the same Helm SDK client-only rendering. The approach described in this RFC follows the same pattern, with the only difference being that charts are embedded in the operator image at build time rather than extracted from component images via init containers at runtime.
+
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
@@ -65,7 +70,7 @@ The kuadrant-operator becomes an umbrella operator that deploys component contro
 
 ## After
 
-1 OLM operator (kuadrant-operator) with a single CSV, bundle, and catalog entry. Component controllers (authorino-operator, limitador-operator, dns-operator, mcp-gateway) are no longer OLM packages. They become internal controller deployments whose lifecycle is managed by the kuadrant-operator. No component controller bundles, catalogs, or CSVs need to be maintained.
+1 OLM operator (kuadrant-operator) with a single CSV, bundle, and catalog entry. Component controllers (authorino-operator, limitador-operator, dns-operator, mcp-gateway) are no longer OLM packages. They become internal controller deployments whose lifecycle is managed by the kuadrant-operator at runtime via Helm chart rendering. No component controller bundles, catalogs, or CSVs need to be maintained.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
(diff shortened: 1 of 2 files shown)


rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,237 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+|---|---|
+| Component controller Deployments | authorino-operator, limitador-operator, dns-operator, mcp-gateway |
+| Component ServiceAccounts | Per-component identity |
+| Component ClusterRoleBindings | Binds component SAs to pre-installed ClusterRoles using `bind`/`escalate` |
+| Component Roles, RoleBindings | Namespace-scoped RBAC (e.g. leader election) |
+| Component Services, ConfigMaps | Metrics, configuration |
+| Wrapper CRs | Authorino CR, Limitador CR (reconciled by child operators into workloads) |
 
-Each step is idempotent — if the umbrella operator restarts mid-migration, it re-checks state and resumes.
+The operator does not create or modify cluster-scoped resources (CRDs, ClusterRoles) at runtime. If a rendered chart produces a ClusterRole or CRD, the operator skips it with a log warning.
```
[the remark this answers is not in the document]
mikenairn: ClusterRoleBindings are explicitly listed in the runtime-managed table. The 'cluster-scoped vs namespaced' distinction is about CRDs and ClusterRoles (installed before the operator starts) vs everything else (created at runtime). ClusterRoleBindings are created at runtime using bind/escalate, which is documented.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,237 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+- **Helm**: Single `helm install` deploys everything. The kuadrant-operator Helm chart includes component CRDs and ClusterRoles in its templates (generated by `kustomize build` from `config/child-operators/crds/` and `config/child-operators/rbac/` at `make helm-build` time).
+- **OLM**: Single operator in the catalog. The bundle includes all component CRDs and ClusterRoles (from `config/child-operators/` via `config/manifests/kustomization.yaml`). No component controller bundles in the catalog.
 
-1. OLM upgrades the umbrella operator bundle — this updates CRDs, ClusterRoles, and ClusterRoleBindings.
-2. The umbrella operator re-renders charts, detects drift, and applies changes in order:
-   - **Workloads** (Authorino, Limitador) and **dns-operator** — updated first, waited on until healthy.
-   - **kuadrant-operator** last — as the policy reconciler, it may depend on new capabilities from the updated workloads.
+Both paths produce the same cluster state: CRDs and ClusterRoles installed, kuadrant-operator running, no component controllers until a Kuadrant CR is created.
```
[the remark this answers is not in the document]
mikenairn: This is an intentional and documented behaviour change. The Guide-level explanation notes that dns-operator previously ran independently. Migration considerations cover resource naming and data plane continuity.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,237 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
-#### Control plane gap
+### RBAC
 
-There is a brief window per component (between operator scale-down and Helm Deployment apply) where no controller is watching that component's configuration. During this window:
+The kuadrant-operator does not need to duplicate component controller permissions. The Kubernetes `bind` verb allows creating ClusterRoleBindings to named ClusterRoles without holding all their permissions. The operator's ClusterRole includes `bind`/`escalate` on each component's ClusterRole `resourceNames`.
 
-- Existing policies continue to be enforced by the running operands
-- New policy changes or CR updates are not reconciled until the replacement is in place
-- No user action is required — the umbrella operator handles the entire sequence
+Only ClusterRole names need tracking, not their contents. When a component controller changes its RBAC rules, no kuadrant-operator change is needed unless a ClusterRole is added or renamed.
```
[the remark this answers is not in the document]
mikenairn: escalate is justified for migration scenarios where the operator may need to modify existing ClusterRoles (e.g. updating ownership annotations/labels during the transition from OLM-managed to umbrella-managed resources). bind alone only covers creating ClusterRoleBindings to existing roles, not modifying the roles themselves.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,268 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+| `RELATED_IMAGE_AUTHORINO` | `quay.io/kuadrant/authorino:latest` | Authorino workload |
+| `RELATED_IMAGE_LIMITADOR_OPERATOR` | `quay.io/kuadrant/limitador-operator:latest` | Limitador operator controller |
+| `RELATED_IMAGE_LIMITADOR` | `quay.io/kuadrant/limitador:latest` | Limitador workload |
+| `RELATED_IMAGE_DNS_OPERATOR` | `quay.io/kuadrant/dns-operator:latest` | DNS operator controller |
+| `RELATED_IMAGE_MCP_GATEWAY` | `ghcr.io/kuadrant/mcp-controller:latest` | MCP Gateway controller |
 
-> Note: the first version of the umbrella operator will contain migration-specific logic that can be removed in subsequent versions.
+These are declared as environment variables on the kuadrant-operator Deployment (`config/manager/manager.yaml`). The downstream build system overrides them with version-pinned or mirrored image references. The Helm reconcilers read these env vars and apply the correct images when rendering child operator charts, either by patching the rendered Deployment image directly (for simple charts that don't support values-based overrides) or by passing Helm values (for charts with configurable image fields).
```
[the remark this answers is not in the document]
mikenairn: The current release process pins these at release time to release specific immutable tags, these are kept pointing to "latest" on main for development.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,268 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
 
-**Test environment:** OpenShift with OLMv0, Kuadrant installed via OLMv0 Subscriptions, a Gateway with HTTPRoute/AuthPolicy/RateLimitPolicy applied, and active ingress traffic.
+The same pattern should be adopted for the kuadrant-operator. When a child repo's CI completes successfully on a tracked branch, it dispatches to the kuadrant-operator. The sync tool runs, and if there are changes, a PR is created or updated. The automation should manage a single open PR per child component (e.g. `sync/authorino-operator`), force-updating it on each trigger rather than accumulating stale PRs.
 
-A background process sends requests through the gateway continuously across all phases, recording success/failure rates.
+This automation must support not just `main` but also `release-*` branches. Multiple kuadrant-operator release branches may track different branches of the same child component. For example, `kuadrant-operator/main` may track `authorino-operator/main` while `kuadrant-operator/release-v1.5` tracks `authorino-operator/release-v1.5`.
```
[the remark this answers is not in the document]
mikenairn: This was updated

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,268 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
 
-| Spec field | Action | Category |
-|---|---|---|
-| (existence of CR) | Previously created Authorino and Limitador operand CRs — now triggers Helm-rendered workload configuration | Operand management |
-| `spec.mtls.*` | Configures PeerAuthentication for mTLS between gateway and operands | Operand configuration |
-| `spec.observability.enable` | Creates ServiceMonitor and PodMonitor CRs | Infrastructure wiring |
-| `spec.observability.tracing.*` | Sets tracing endpoint on Authorino and Limitador | Operand configuration |
-| `spec.observability.dataPlane.*` | Drives wasm-shim configuration via effective policy computation | Data plane configuration |
-| `spec.components.developerPortal.enabled` | Creates/deletes Developer Portal Deployment | Operand management |
+`bind` is sufficient for steady-state runtime operation (creating ClusterRoleBindings). `escalate` is retained for migration scenarios where the operator may need to update existing ClusterRole rules to align with the consolidated model. Whether rule changes are needed depends on the migration implementation. If migration only requires metadata updates (annotations, labels), `escalate` can be removed.
```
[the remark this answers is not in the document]
mikenairn: removed escalate

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,368 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
-### Umbrella Operator
+| Resource | Source | Notes |
+|---|---|---|
+| All CRDs (kuadrant + component) | `config/child-operators/crds/` and `config/crd/` | Installed before the operator starts |
+| Component ClusterRoles | `config/child-operators/rbac/` | Installed before the operator starts |
+| kuadrant-operator Deployment | `config/manager/` | The operator itself |
+| kuadrant-operator ServiceAccount, ClusterRole, ClusterRoleBinding | `config/rbac/` | Operator's own RBAC |
 
-A single operator published as an OLMv0 package with zero dependency declarations. It manages all Kuadrant components internally by rendering Helm charts at runtime using Helm's Go SDK with `ClientOnly=true` and `DryRun=true` — pure template rendering with no Helm release tracking. Rendered manifests are applied via controller-runtime. Charts are embedded in the umbrella operator image at build time.
+**Managed by the kuadrant-operator at runtime (triggered by Kuadrant CR creation):**
```
eguzki: This RFC proposes that component ServiceAccounts, ClusterRoleBindings, and namespace-scoped Roles/RoleBindings be created dynamically by kuadrant-operator at runtime, using the `bind` verb against pre-installed ClusterRoles. This requires kuadrant-operator to hold `bind` permission and track exact ClusterRole `resourceNames`, and it's called out as a drawback. 

kuadrant-operator already solves this exact problem for `developer-portal-controller` without any of that machinery, and the mechanism is worth considering here.

- `config/dependencies/developer-portal/kustomization.yaml` vendors the child's `service_account.yaml`, `role.yaml`, `role_binding.yaml`, and leader-election Role/RoleBinding from upstream, with `namePrefix: developer-portal-`
- `kustomize build config/manifests | operator-sdk generate bundle ...` then converts these raw RBAC objects into CSV form. The `--extra-service-accounts=developer-portal-controller-manager` bundle-gen flag is what tells `operator-sdk` to track this as a legitimate second SA rather than assuming a single-SA operator built only around `kuadrant-operator-controller-manager`.
- At install time, OLM reads those permission lists and provisions the ServiceAccount plus Role/ClusterRole plus binding for every distinct `serviceAccountName` itself. The kuadrant-operator never creates a ServiceAccount object programmatically. 

Applying the same pattern to authorino-operator, limitador-operator, dns-operator, and mcp-gateway is worth considering here. 

Helm charts also include those extra SA, RO and CRO using some kustomization patches specifically for the helm manifests.

eguzki: @Patryk-Stefanski knows about this even better than myself.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,368 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+- **Removing intermediate operator layers and wrapper CRs**: deploying Authorino and Limitador workloads directly from the kuadrant-operator, eliminating authorino-operator and limitador-operator as running controllers. This removes the `Authorino` and `Limitador` CRDs from the user-facing API surface. Each operator removed is a separate container image with its own Go dependency tree. Every transitive dependency is a potential CVE that needs scanning, patching, and releasing. Removing these controllers reduces the security surface area and maintenance burden.
+- **Extracting a dedicated policy controller**: separating policy reconciliation (AuthPolicy, RateLimitPolicy, DNSPolicy, TLSPolicy) from the kuadrant-operator into a `kuadrant-policy-controller` deployed as a component, making the kuadrant-operator purely an orchestration layer.
```
eguzki, line 242: This brings peace to my mind. Calling things by their own name.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,368 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+- **Removing intermediate operator layers and wrapper CRs**: deploying Authorino and Limitador workloads directly from the kuadrant-operator, eliminating authorino-operator and limitador-operator as running controllers. This removes the `Authorino` and `Limitador` CRDs from the user-facing API surface. Each operator removed is a separate container image with its own Go dependency tree. Every transitive dependency is a potential CVE that needs scanning, patching, and releasing. Removing these controllers reduces the security surface area and maintenance burden.
```
eguzki, line 239: This is in my opinion the end goal as we understand kuadrant today. One operator to rule them all. Just one package being delivered to be installed if you want to use any kuadrant component. No matter upstream or downstream.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: Yes, i did look at this when investigating during the POC. This however i believe would limit us to having the controllers run inside the same namespace as the kuadrant operator itself. Currently, maybe not described fully in this RFC, the control plane pieces (child controllers) will be deployed in the same namespace as the kuadrant CR moving us towards more of an ability to create multi-tenanted control planes (Multiple KuadrantCRs).

eguzki (approved):
This looks good to me. It's a great piece of design work that paints the picture of a big challenge and a lot of work to be done.

IMHO, it's a stepping stone toward a single-operator model that governs the lifecycle of Kuadrant (including all its components).

I left a few comments to consider or think about.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: That said, maybe these should be deployed in the same namespace :thinking:

eguzki: > This however i believe would limit us to having the controllers run inside the same namespace as the kuadrant operator itself.

I believe so too.

I do not think that deploying child operators in the same namespace as the kuadrant operator itself impacts on the multiple kuadrant CRs user story, though. But this is another discussion.

maleck13: I think it is fine to deploy the control plane operators / pieces into the same namesapce as the Kuadrant Operator as long as they are cluster scoped. Really when we talk about multiple kuadrant CRs (as mentioned different discussion) it shouldn't trigger a new set of control plane pieces but rather a new set of data plane components (Authorino, Limitador). Allowing an isolated data plane deployment is what would be important. Whatever we do here, shouldn't block that as an option in the future

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,372 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
 
-This follows the pattern used by OpenShift's [cluster-olm-operator](https://github.com/openshift/cluster-olm-operator).
+## Runtime rendering
 
-### Helm Chart Refactoring
+When a user creates a Kuadrant CR, the kuadrant-operator renders each component controller's Helm chart from `/charts/<name>/` in the container image using the Helm Go SDK (`ClientOnly=true`, `DryRun=true`). The rendered namespaced resources are applied via Server-Side Apply with the kuadrant-operator as the field manager.
 
-The current kuadrant-operator chart (`charts/kuadrant-operator/`) is a monolithic kustomize-generated blob — a single `manifests.yaml` of ~14,500 lines produced by `kustomize build config/helm`. This must be refactored to support per-component rendering, ordered deployment, and direct workload deployment.
+Component controllers are deployed into the Kuadrant CR's namespace, not the kuadrant-operator's namespace. This is a change from the current model where all child operators run in `kuadrant-system` regardless of the Kuadrant CR's location. This approach moves towards supporting future multi-tenanted control plane scenarios.
```
maleck13: As mentioned above not sure this matters so much for control plane components. Really it is the Authorino and Limitador components that should be installed in the same ns as the Kuadrant CR (which in most cases will currently be the kuadrant-system namespace) so it is the data plane.

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,372 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+This automation must support not just `main` but also `release-*` branches. Multiple kuadrant-operator release branches may track different branches of the same child component. For example, `kuadrant-operator/main` may track `authorino-operator/main` while `kuadrant-operator/release-v1.5` tracks `authorino-operator/release-v1.5`.
 
-1. OLM upgrades the umbrella operator bundle — this updates CRDs, ClusterRoles, and ClusterRoleBindings.
-2. The umbrella operator re-renders charts, detects drift, and applies changes in order:
-   - **Workloads** (Authorino, Limitador) and **dns-operator** — updated first, waited on until healthy.
-   - **kuadrant-operator** last — as the policy reconciler, it may depend on new capabilities from the updated workloads.
+Changes to the release process and version pinning strategy are out of scope for this RFC and will be addressed separately.
 
-### Version Pinning
+## Migration from multi-operator OLM installation
```
maleck13, line 202: Not a migration question. Do you anticipate any changes to the local developer setup around this work? Same question for CI e2e setup. Can each of these potentially exercise either the OLM or Helm path?

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,372 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
+- Create a Kuadrant CR with active policies and ingress traffic
+- Perform pre-upgrade cleanup (remove child subscriptions and CSVs)
+- Upgrade to the consolidated release via OLM catalog channel
+- Verify all child CRDs survived the migration
+- Verify kuadrant-operator CSV declares ownership of all CRDs
+- Verify all component controllers are running (now managed by kuadrant-operator)
+- Verify zero data plane disruption throughout (policies continue to be enforced, no traffic loss)
+- Verify CRD schema changes are applied correctly when present
+
+These tests should run in CI for every release and serve as ongoing regression testing for the OLM upgrade path.
```
maleck13, line 232: Yes I would have thought we already have something like this in place minus the migration pieces. If not we definitely need this in place and it should be executed for every release

maleck13 (approved):
Generally approve of this more focused approach . Only real comments here are that the multi-tenancy piece is really for the data plane only. MCP-gateway already supports that and have a shared control plane is fine

rfcs/0019-olmv1-operator-consolidation.md
```diff
@@ -23,292 +40,372 @@ Bundles declaring `olm.package.required`, `olm.gvk.required`, or `olm.constraint
...
-### Day 2 Considerations
+The `ResolutionFailed` condition message clearly identifies the GVK conflict, making it straightforward to diagnose and point users at the migration documentation.
 
-The dependency operators currently handle day 2 operations that must have a home in the new architecture.
+### Alternatives considered for migration
 
-**PodDisruptionBudgets:** limitador-operator optionally creates a PDB for Limitador when `spec.pdb` is set (not created by default). authorino-operator does not manage a PDB. In the new architecture, users create PDBs directly if needed — no operator involvement required.
+#### Transition release with runtime CRD management
 
-**Scaling replicas:** Both operators support configuring replicas via their CRs. In the new architecture, initial replica counts are set via Helm values at install time. Users can scale workloads directly via `kubectl scale` — since the umbrella operator owns the Deployments (not OLM), user-applied scaling is not reverted. The umbrella operator must respect user-set replica counts and not overwrite them on reconciliation.
```
guicassolato, line 236: I know this line is being removed, but still leaving a nit here for the record, once/if management of component deployments directly by the umbrella operator returns to scope.

In order to scale Authorino up from 1, apart from `kubectl scale`, users must also add `--enable-leader-election` to the command-line arguments of the Authorino pod template. Today, this is managed by the Authorino Operator. It is a particular way of mutating the `Authorino` CR (from `replicas: 1` as well as back to `replicas: 1`) that causes redeployment of all Authorino pods.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: This is out of date now, the control plane components are no longer installed in the Kuadrant CR namespace.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: I'd say we do, this is just highlighting that we need to ensure that upgrades are tested thoroughly considering some of the changes we will be making. I'd also kind of like to see a test like that executed continuously, maybe on merges to main, but also optionally executable on PRs that we think might require an upgrade test.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: Good to know. Thanks.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: The RFC has been updated to use runtime Helm chart rendering rather than the split chart approach. Child operator ClusterRoles and SAs are now created at runtime from the chart (using escalate and bind), so the kustomize vendoring pattern is no longer needed.

mikenairn: Control plane operators deploy in the operator namespace. Data plane (Authorino/Limitador workloads) deploy in the Kuadrant CR namespace.

rfcs/0019-olmv1-operator-consolidation.md
mikenairn: There should be minimal changes to local dev setup. The separate deploy-dependencies step that applies child operator manifests before running the kuadrant-operator will be removed. The operator handles this at startup now, so it may take slightly longer to initialise.

The current dev and e2e setup don't use OLM or Helm installation paths directly, they deploy the controller via kustomize. Both OLM and Helm paths use the same runtime rendering, so there's no divergent behaviour to test separately. CI e2e tests exercise whichever gateway provider is configured, same as today.

guicassolato (commented):
One late approval here echoing on the positive comments left before.

A few minor things to have in mind during implementation:

1. **RESTMapper cache staleness** — make sure the client used to verify the presence of the CRDs after install has a clean vision of the state. The new pre-startup step should verify CRDs appear in discovery, not just that the apply succeeded.

2. **No retry mechanism** — CRD presence is checked once at boot. If the operator's own Helm-chart-apply step is slow or partially fails, child controllers are silently disabled with no recovery path short of pod restart. Need to decide whether to block operator startup (with timeout) or treat partial failure as degraded.

3. **Failure semantics change** — today, "Limitador CRD missing" means "admin chose not to install Limitador." Under the RFC, it means "my own startup failed." The operator should surface this distinction, e.g. via some special status condition on the Kuadrant CR perhaps.

PR #190, opened 2026-07-21
RFC 0021: Token rate limit reservations

eguzki:
## Summary by CodeRabbit

* **Documentation**
  * Added a new RFC describing token rate-limit reservations to prevent capacity race conditions.
* **New Features**
  * Introduced a cluster-wide reservation mode for token rate limiting to reserve capacity on request arrival and commit usage after response completion.
  * Added optional per-limit reservation settings (reservation amount and expiry) plus safety controls to cap reservation TTL and fraction.
  * Included rollback/compatibility controls to retain legacy Check/Report behaviour where required.


rfcs/0000-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,299 @@
...
+- **the distributed gossip-based backend**: the reservation registry is also local-memory-only for this RFC, but here that's a deliberate scope choice, not a backend limitation. Reservations are therefore only guaranteed within a single replica on this backend for now. Future RFC extends the gossip protocol to cover reservations too (see [Future possibilities](#future-possibilities)).
+
+This distinction is surfaced to operators (docs and/or a status condition) rather than left implicit, since it changes what "the race is closed" means depending on backend and replica count.
+
+### Limitador-level opt-in
+
+A server-level flag, `--enable-reservations` (default: `false`, i.e. reservations disabled), independent of anything the Kuadrant CR requests. When not set, `Reserve`/`Commit` are not registered (or return `UNIMPLEMENTED`). This exists as defense-in-depth for operators who want a hard kill-switch regardless of what any CR asks for.
```
maksymvavilov: The Kuadrant default is to enable, and the Limitador default is to disable? 
`--enable-reservations` here vs `--disable-reservations` a few lines below vs `disable: false ` in kuadrant CR . 
Even if the CR will have an enum of: `strategy: reservation/default` (or something like this, as discussed at the community call) we might want the same "default" state across all components

rfcs/0000-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,299 @@
...
+
+### TTL and expiry
+
+A reservation's `expires_at`, computed independently **for each counter it touches**, is:
+
+```text
+expires_at(counter) = min(
+    now + reservation.ttl,           // caller-supplied TTL (CEL-computed, see above)
+    current_window_expiry(counter),  // this counter's own window boundary
+    server_max_reservation_ttl       // operator-configured safety clamp
```
maksymvavilov: How do we set the `server_max_reservation_ttl`?

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,299 @@
...
+                                    // counter this reservation touched, so one
+                                    // token released by the caller settles all
+                                    // of them in a single Commit call
+    amount:         u64            // estimated tokens reserved, uniform across
+                                    // every counter this reservation touched
+    expires_at:     Timestamp      // see "TTL and expiry" below
+}
+
+// Capacity available for a counter must now account for outstanding holds:
+outstanding(counter) = sum(e.amount for e in registry[counter] if e.expires_at > now)
```
maksymvavilov, line 166: This might be nothing, but this will loop over every reservation/request in process (O(n) for n reservations). 
Is keeping a single counter that each reservation will update a better idea (O(1))?

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,299 @@
...
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+- Original spec: [TokenRateLimitPolicy GA](https://docs.google.com/document/d/1Lmy52cZZuteJj8cxVNEcC5q59YYB2MV74naGKpkweds/edit?tab=t.0#heading=h.803yw4a9qwi9)
+
+# Summary
+
+`TokenRateLimitPolicy` ([RFC 0013](./0013-ai-policies.md)) checks capacity when a request arrives but only reports actual token usage after the model responds, which lets concurrent in-flight requests race past a limit before any of them report usage. This RFC introduces a **reservation layer** in Limitador: on arrival, an estimated token volume is reserved and held (with a TTL) against a limit's capacity; when the response comes back, the caller commits the reservation with actual usage in a single call, releasing any over-reservation immediately. Rejection semantics, rate-limit headers, and multi-limit all-or-nothing behavior are unchanged. The feature is additive at every layer (new gRPC RPCs, new CRD field, new Limitador flag) and is enabled by default with an explicit opt-out.
+
+# Motivation
```
maksymvavilov, line 13: I was looking at it from the Egress perspective: adding the Egress gateway as motivation will make the case for this much stronger (not that it needs that, but hey, why not?)

Also, for the egress - we will be making a heavy guess on the reservation size. Maybe the default of `100` is too generous? I'd rather over-reserve with ~5k or something like that. What do you think (make it configurable in kuadrant CR?)?  
I'm talking here about the default that is hardcoded [here](https://github.com/Kuadrant/architecture/pull/190/changes#diff-d4b11b4cbbf6542ccca7b0fdac252ea22d1aeb6fdcd85c0a6ca80496ac8f6ad6R87-R90)

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,317 @@
...
+Nothing has to change. On a cluster running a compatible Limitador and operator, every `TokenRateLimitPolicy` automatically starts holding capacity for the estimated token cost of a request the moment it arrives, instead of only accounting for it after the model responds. No CR edits or policy edits required to get this protection.
+
+The one new thing available is a **cluster-wide mode switch**, on the `Kuadrant` CR:
+
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+spec:
+  tokenRateLimiting:
+    mode: Reservation   # default; set to CheckReport to revert every
```
maleck13, line 46: Yes ok this feels better than a kill switch. I don't know what reason someone would chose the old behaviour though, but if it may be the case we add new "modes" then a choice is useful.

rfcs/0000-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,317 @@
...
+- **the distributed gossip-based backend**: the reservation registry is also local-memory-only for this RFC, but here that's a deliberate scope choice, not a backend limitation. Reservations are therefore only guaranteed within a single replica on this backend for now. Future RFC extends the gossip protocol to cover reservations too (see [Future possibilities](#future-possibilities)).
+
+This distinction is surfaced to operators (docs and/or a status condition) rather than left implicit, since it changes what "the race is closed" means depending on backend and replica count.
+
+### Limitador-level opt-in
+
+A server-level flag, `--enable-reservations` (default: `false`, i.e. reservations disabled), independent of anything the Kuadrant CR requests. When not set, `Reserve`/`Commit` are not registered (or return `UNIMPLEMENTED`). This exists as defense-in-depth for operators who want a hard kill-switch regardless of what any CR asks for.
```
maleck13: interesting so I wonder is this enough and so can avoid the additional API in the kuadrant CR? This boils  down to the same type of operation. Patch a resource with a new value?

maleck13 (approved):
This looks good from my perspective. I only question the need for the Kuadrant level api to set the mode. If we have a "bail out option" on limitador do we really need another one on the kuadarant CR? Its not completely clear to me when a user would actually want the old behaviour to come back beyond some catostrophic bug?

That said I don't think this is a blocker it just sets a precendent where more and more policy level behaviour could be added to the Kuadrant CR.

Generally approve of the changes outlined though

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,317 @@
...
+- Issue tracking: ~[Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)~
+- Original spec: [TokenRateLimitPolicy GA](https://docs.google.com/document/d/1Lmy52cZZuteJj8cxVNEcC5q59YYB2MV74naGKpkweds/edit?tab=t.0#heading=h.803yw4a9qwi9)
+
+# Summary
+
+`TokenRateLimitPolicy` ([RFC 0013](./0013-ai-policies.md)) checks capacity when a request arrives but only reports actual token usage after the model responds, which lets concurrent in-flight requests race past a limit before any of them report usage. This RFC introduces a **reservation layer** in Limitador: on arrival, an estimated token volume is reserved and held (with a TTL) against a limit's capacity; when the response comes back, the caller commits the reservation with actual usage in a single call, releasing any over-reservation immediately. Rejection semantics, rate-limit headers, and multi-limit all-or-nothing behavior are unchanged. The feature is additive at every layer (new gRPC RPCs, new CRD field, new Limitador flag) and is enabled by default, with an explicit `CheckReport` mode to revert to today's behavior.
```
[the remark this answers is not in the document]
eguzki: That should not be in summary. Mentioned in storage section

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,317 @@
...
+                                    // token released by the caller settles all
+                                    // of them in a single Commit call
+    amount:         u64            // estimated tokens reserved, uniform across
+                                    // every counter this reservation touched
+    expires_at:     Timestamp      // see "TTL and expiry" below
+}
+
+// Capacity available for a counter must now account for outstanding holds:
+outstanding(counter) = sum(e.amount for e in registry[counter] if e.expires_at > now)
+admit if counter.value + outstanding(counter) + requested_amount <= limit
```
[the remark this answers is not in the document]
eguzki: implementation detail. Reserve keeps the all-or-nothing behavior is announced and that suffices.

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,317 @@
...
+  string reservation_id = 3;                // optional; empty behaves like plain Report
+  uint64 actual_amount = 4;
+}
+
+message CommitResponse {
+  bool reservation_released = 1; // diagnostic only: false if reservation_id
+                                  // was empty, unknown, or already expired,
+                                  // and actual_amount was applied directly
+                                  // instead (Report-style) -- see below
+}
```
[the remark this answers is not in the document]
eguzki: That's correct, expected and an acceptable trade-off. It's not a new risk, it's inherited, unchanged behavior (report call). A late or lost reservation degrades gracefully to today's exact behavior.

rfcs/0000-token-rate-limit-reservations.md
eguzki: good catch. Fixing this

rfcs/0000-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,341 @@
...
+
+```
+has(requestBodyJSON('/max_tokens'))
+  ? uint(requestBodyJSON('/max_tokens'))
+  : 100
+```
+
+`max_tokens` is the client's own ceiling on completion length, so this default is a real, protocol-declared upper bound rather than a heuristic estimate, whenever the client happens to supply one.
+
+Note that the `amount` default assumes `requestBodyJSON('/max_tokens')` is reliably readable at this point, which is itself unresolved (see [Unresolved questions](#unresolved-questions)). If it isn't, this default degrades to the flat fallback constant for every request.
```
adam-cattermole: Yep - I realise this is noted here, but the reading of the request body is currently unsupported due to the issue found with Envoy's `allow_on_headers_stop_iteration` when used in conjunction with a wasm filter and subsequent filters - so it's likely we'll end up with the fallback for the time being.

rfcs/0000-token-rate-limit-reservations.md
eguzki: On the Limitador CR which is passed to the limitador server.

eguzki: adding this to the rfc

rfcs/0021-token-rate-limit-reservations.md
```diff
@@ -0,0 +1,341 @@
...
+
+```text
+expires_at(counter) = min(
+    now + reservation.ttl,           // caller-supplied TTL (CEL-computed, see above)
+    current_window_expiry(counter),  // this counter's own window boundary
+    server_max_reservation_ttl       // operator-configured safety clamp
+)
+```
+
+Clamping to the counter's own window boundary is what prevents a reservation from bleeding into the *next* window if the model call outlives the current window. A counter with no prior writes gets its window initialized by `Reserve`.
```
adam-cattermole, line 184: Do I understand correctly that this mean the `Reservation` is removed when the counter resets? does this mean in the edge case that I could have made a reservation, but someone else consumed all of the quota before my upstream responded and I `Commit` my usage?

rfcs/0021-token-rate-limit-reservations.md
eguzki: The O(n) is real, but bounded by something narrower than it looks. The case where this stops being true is a coarse, shared counter under high concurrency, where thousands of requests could be outstanding against the same counter simultaneously. For the common case, a counter keyed per identity, e.g. auth.identity.userid, one user rarely has more than a handful of requests in flight at once. 

Furthermore, the per-entry list isn't just serving the sum. It's also what Commit looks up by reservation_id to remove. A single aggregate would need atomic decrement on whichever happens first, commit or expiry. Today's design avoids that.

Makes sense?

rfcs/0021-token-rate-limit-reservations.md
eguzki: Thanks for both points.

Egress motivation added: longer/variable RTT to an external upstream widens the race window

Default size: On making it configurable at the Kuadrant CR level: not adding that. `reservation.amount` is already CEL on the policy itself, so set your own if the generated default doesn't fit. Having low default, i.e. under-reserving silently defeats the point of this RFC, while over-reserving just costs some concurrency/false throttling until Commit corrects it. I will err toward the higher end of the range you're weighing. I will set 5k.

rfcs/0000-token-rate-limit-reservations.md
eguzki: updated

rfcs/0021-token-rate-limit-reservations.md
eguzki: > Do I understand correctly that this mean the Reservation is removed when the counter resets?

Yes, that's correct, and it's not new to this RFC. It's the same behavior `Report` already has today.

> does this mean in the edge case that I could have made a reservation, but someone else consumed all of the quota before my upstream responded and I Commit my usage?

`Commit` (like `Report`) re-resolves counters using current time and applies `actual_amount` unconditionally, with no re-admission check. If your request straddles a window boundary, your reservation (clamped to `current_window_expiry`) expires at that boundary, and by the time you `Commit`, you're landing in the *new* window's counter. The `new` window counter which your original `Reserve` never protected and which could already be fully consumed by other traffic.

adam-cattermole: Gotcha yeah that makes perfect sense - just making sure I understand the exact behaviour - so in the case where you only consume your reservation or less you're limited to 2x counter in this edge case across boundaries (entire limit is reserved and not committed until next window where it's fully reserved again and all are committed together). Obviously if you reserve less than you consume then it's worse, but no worse than today

maleck13 (approved):

PR #191, opened 2026-07-24
RFC: Token rate limit soft/hard boundaries

jland-redhat:
## Summary

- Draft RFC proposing **soft** vs **hard** boundaries on `TokenRateLimitPolicy`: soft exceed admits traffic and injects configurable headers (name + value); hard exceed remains `429`.
- Asks that **all matching limits** be evaluated so multiple soft signals can fire on one request (e.g. personal soft → deprioritize / smaller model while still metering against a team budget).
- Change-request style: illustrative API shape only; Limitador/wasm-shim/CRD design left to Kuadrant.

## Motivation / context

Follow-up from discussion with Craig / Eguzki after [architecture#190](https://github.com/Kuadrant/architecture/pull/190). Soft/hard framing from the community call.

## Test plan

- [ ] Kuadrant review of the ask (soft headers + multi-bucket evaluation + hard ceiling)
- [ ] Align on open questions in the RFC (request vs response headers, defaults, CEL, etc.)
- [ ] After consensus / FCP: assign RFC number and track implementation issues

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC proposing soft and hard token rate-limit boundaries.
  * Documents configurable headers for soft-limit signals and `429` responses for hard-limit enforcement.
  * Defines evaluation behaviour for multiple matching limits and preserves hard-limit behaviour by default.
  * Includes configuration examples, expected outcomes, rationale, alternatives, unresolved questions and success criteria.

### DIFF
--- a/rfcs/0000-token-rate-limit-soft-hard-boundaries.md
+++ b/rfcs/0000-token-rate-limit-soft-hard-boundaries.md
@@ -1,0 +1,194 @@
+# Token Rate Limit Soft / Hard Boundaries
+
+- Feature Name: `token_rate_limit_soft_hard_boundaries`
+- Status: Draft
+- Start Date: 2026-07-24
+- RFC PR: (to be filled when opened against [Kuadrant/architecture](https://github.com/Kuadrant/architecture))
+- Issue tracking: (to be filled)
+- Authors: James Land (MaaS / RHOAI), with input from Eguzki Astiz Lezaun / Craig Brookes (Kuadrant)
+- Related: [RFC 0013 — AI Policies / `TokenRateLimitPolicy`](./0013-ai-policies.md)
+- Context: Soft/hard framing from Kuadrant community call (idea attributed to Alex); follow-up with MaaS
+
+# Summary
+[summary]: #summary
+
+Today `TokenRateLimitPolicy` treats every limit as a **hard** gate: when a counter is exhausted the request is rejected with `429`. That is necessary for capacity protection, but it is too blunt for real user use cases. Customers need **flexible business logic** after a budget is crossed — deprioritize traffic, route to a cheaper model, keep metering against a team pool, degrade quality — without Kuadrant encoding every product policy in the rate-limit plane.
+
+**Ask:** support **soft and hard boundaries** on token rate limits:
+
+- **Soft boundary exceeded** → request is **still admitted**, and the gateway injects **configurable headers** (both key and value under policy control) marking which limit(s) were exceeded.
+- **Hard boundary exceeded** → request is **rejected** with `429` (today’s behavior).
+
+**Every matching limit should be evaluated**, not fail-fast on the first over-limit. One request may hit multiple soft buckets (and possibly a hard one), producing multiple headers or multiple values. Downstream plugins own what to do with those signals; Kuadrant owns metering and signaling.
+
+# Motivation
+[motivation]: #motivation
+
+Conversations with customers make one point clear: **flexibility matters**. Hard “deny with 429” cannot cover every enforcement story:
+
+| Need (examples) | Hard 429 alone |
+|---|---|
+| Over personal quota → deprioritize / smaller model, still count against team | Blocks traffic that should continue under different handling |
+| Over soft SLA → mark non-priority / lower QoS | Cannot signal; only deny or allow |
+| Over budget → route to a cheaper model | Needs a signal, not a reject |
+
+Hard limiting should remain available as a **safety ceiling**. Soft boundaries separate *“did this request cross a budget?”* from *“what should we do about it?”*.
+
+## Soft / hard framing
+
+Aligned with the community call discussion:
+
+1. Cross the **soft** boundary → traffic continues, with classification / rate-limit headers attached.
+2. Cross the **hard** boundary → traffic is rejected with `429`.
+
+Existing informational headers (`X-RateLimit-Limit`, `X-RateLimit-Remaining`, `X-RateLimit-Reset`) stay useful. We also need **policy-authored headers** (configurable name + value) so platforms can drive downstream behavior.
+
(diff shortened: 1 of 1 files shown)


rfcs/0000-token-rate-limit-soft-hard-boundaries.md
```diff
@@ -0,0 +1,194 @@
...
+
+# Summary
+[summary]: #summary
+
+Today `TokenRateLimitPolicy` treats every limit as a **hard** gate: when a counter is exhausted the request is rejected with `429`. That is necessary for capacity protection, but it is too blunt for real user use cases. Customers need **flexible business logic** after a budget is crossed — deprioritize traffic, route to a cheaper model, keep metering against a team pool, degrade quality — without Kuadrant encoding every product policy in the rate-limit plane.
+
+**Ask:** support **soft and hard boundaries** on token rate limits:
+
+- **Soft boundary exceeded** → request is **still admitted**, and the gateway injects **configurable headers** (both key and value under policy control) marking which limit(s) were exceeded.
+- **Hard boundary exceeded** → request is **rejected** with `429` (today’s behavior).
```
adam-cattermole, line 20: I think this is a great candidate for the extensions SDK - this kind of custom behaviour _should_ be something we support and I think this would be an interesting 'trial' to put the extensions through its paces to see if there's anything we're missing

rfcs/0000-token-rate-limit-soft-hard-boundaries.md
```diff
@@ -0,0 +1,194 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+Soft/hard quotas appear in batch/HPC schedulers and cloud “burst” SKUs. Many API gateways emit `X-RateLimit-*` while still serving (client-driven soft). This proposal is closer to *server-side classification*: continue serving, but mark the request for downstream policy.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Request headers vs response headers (or both) for soft signals?
```
eguzki, line 175: Worth noting this is a pure product-design choice, not a technical constraint, both are feasible with a single evaluation.Since the soft/hard boundary is checked against already-accumulated counter state (from prior requests), the decision is known at the request-headers phase, before the request proxies upstream and before any body inspection. 

No re-evaluation or second Limitador round-trip is needed to support both

rfcs/0000-token-rate-limit-soft-hard-boundaries.md
```diff
@@ -0,0 +1,194 @@
...
+[summary]: #summary
+
+Today `TokenRateLimitPolicy` treats every limit as a **hard** gate: when a counter is exhausted the request is rejected with `429`. That is necessary for capacity protection, but it is too blunt for real user use cases. Customers need **flexible business logic** after a budget is crossed — deprioritize traffic, route to a cheaper model, keep metering against a team pool, degrade quality — without Kuadrant encoding every product policy in the rate-limit plane.
+
+**Ask:** support **soft and hard boundaries** on token rate limits:
+
+- **Soft boundary exceeded** → request is **still admitted**, and the gateway injects **configurable headers** (both key and value under policy control) marking which limit(s) were exceeded.
+- **Hard boundary exceeded** → request is **rejected** with `429` (today’s behavior).
+
+**Every matching limit should be evaluated**, not fail-fast on the first over-limit. One request may hit multiple soft buckets (and possibly a hard one), producing multiple headers or multiple values. Downstream plugins own what to do with those signals; Kuadrant owns metering and signaling.
```
eguzki, line 22: I agree on this. 

For future reference, a note on changes requried in limitador to have this working as defined. 

wasm-shim batches all matching limits into a single Limitador `CheckRateLimit` call. Today, that path (a) drops counter data entirely returning empty counters and (b) short-circuits on the first over-limit counter rather than evaluating all matching limits. No per-limit attribution is available today when multiple limits are batched. This will need real changes in Limitador, not just gateway-side interpretation

eguzki (commented):
The overall idea is good and great improvement. 

Consider the alternative schema: `rates` + `softRates` instead of a `boundary` field

Rather than a single `rates` list per limit with a `boundary: Soft|Hard` flag, consider splitting into two optional stanzas per limit: `rates` (hard, unchanged from today) and `softRates` (soft). This avoids duplicating `when`/`counters` across separate limit blocks when you want soft then hard progression on the *same* bucket, and it naturally covers hard-only, soft-only, and soft+hard cases with one shape:

```yaml
limits:
  # Soft threshold, then a hard ceiling on the SAME bucket
  personal-daily:
    when:
    - predicate: 'auth.identity.userid != ""'
    counters:
    - expression: auth.identity.userid
    softRates:
    - limit: 40000
      window: 1d
    softSignal:
      headers:
      - name: "X-Kuadrant-Limit-Exceeded"
        value: "personal-daily"
      - name: "X-MaaS-Traffic-Class"
        value: "non-priority"
    rates:
    - limit: 50000
      window: 1d

  # Soft-only — no hard ceiling on this bucket at all
  team-daily:
    when:
    - predicate: 'auth.identity.organization != ""'
    counters:
    - expression: auth.identity.organization
    softRates:
    - limit: 500000
      window: 1d
    softSignal:
      headers:
      - name: "X-Kuadrant-Limit-Exceeded"
        value: "team-daily"

  # Hard-only — today's behavior, unchanged
  model-capacity:
    rates:
    - limit: 2000000
      window: 1m
    counters:
    - expression: '"model:" + requestBodyJSON("/model")'
```

Notes:
- `boundary: Soft|Hard` goes away. Soft vs. hard is implicit in which stanza a threshold is under.
- Backward compatible: existing policies only ever set `rates`, unchanged.
- CRD validation needs to relax "`rates` required" to "at least one of `rates`/`softRates` required."

PR #196, opened 2026-08-11
 RFC 0022  Runtime-Managed Network Policies for Kuadrant

Boomatang:
# Summary

Introduces RFC 0022, which proposes runtime-managed Kubernetes NetworkPolicy resources for ingress-only network isolation across all Kuadrant components.

The kuadrant-operator will dynamically create and manage standard `networking.k8s.io/v1` NetworkPolicy objects to establish a deny-by-default ingress posture with explicit allow rules for known traffic patterns (metrics scraping, gRPC, wasm-shim serving, OIDC discovery, and console plugin access). This is an always-on feature requiring no user configuration.

# Motivation

Today all Kuadrant pods accept ingress traffic from any source in the cluster with no defense-in-depth at the network layer. This RFC closes that gap by defining a label-scoped deny model that avoids disrupting non-Kuadrant workloads sharing the same namespaces.

# Design Highlights

- **Ingress-only deny model** - egress is left unrestricted to avoid the complexity of tracking dynamic outbound destinations (Redis, OIDC providers, API server)
- **Label-scoped deny** - deny policies target only pods with `kuadrant.io/managed: "true"`, leaving other workloads in shared namespaces unaffected
- **Two-namespace support** - separate policy sets for the operator namespace and the operand namespace (Kuadrant CR namespace), with deduplication when they are the same
- **Dynamic gateway namespace handling** - ingress allow rules for wasm-shim and gRPC ports are updated automatically as Gateway resources appear and disappear across namespaces
- **Standard NetworkPolicy** - uses the GA `networking.k8s.io/v1` API, supported by all CNI plugins, avoiding the alpha AdminNetworkPolicy and its OVN-Kubernetes requirement
- **Owner reference lifecycle** - operator-namespace policies are owned by the operator Deployment; operand-namespace policies are owned by the Kuadrant CR, enabling automatic garbage collection on deletion

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC proposing always-on, runtime-managed Kubernetes NetworkPolicies.
  * Defines scoped ingress-deny and component-specific allow rules for operator and workload namespaces.
  * Documents dynamic Gateway namespace handling and optional console-plugin access.
  * Covers labelling prerequisites, upgrade sequencing, permissions, reconciliation, namespace changes, cleanup, port mappings, alternatives, and open questions.
  * This release documents the proposal; it does not itself enable the policies.

### DIFF
--- a/rfcs/0022-network-policy.md
+++ b/rfcs/0022-network-policy.md
@@ -1,0 +1,569 @@
+# Runtime-Managed Network Policies for Kuadrant
+
+- Feature Name: `network-policy`
+- Status: Draft
+- Start Date: 2026-08-10
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [CONNLINK-1205](https://issues.redhat.com/browse/CONNLINK-1205)
+
+# Summary
+[summary]: #summary
+
+The kuadrant-operator will dynamically create and manage standard Kubernetes NetworkPolicy objects (`networking.k8s.io/v1`) at runtime to provide ingress-only network isolation for all components across the operator and operand namespaces.
+This is an always-on feature requiring no user configuration.
+
+Policies are created in two scopes:
+
+- **Operator namespace**: A label-scoped deny policy targeting pods with `kuadrant.io/managed: "true"`, plus per-operator allow policies.
+- **Operand namespace** (the namespace where the Kuadrant CR is created): A label-scoped deny policy targeting pods with `kuadrant.io/managed: "true"`, plus per-operand allow policies.
+
+Both namespaces use label-scoped deny to avoid disrupting other workloads that may share the namespace.
+When the operator and operand namespaces are the same, a single deny policy and the combined set of allow policies are created.
+
+# Motivation
+[motivation]: #motivation
+
+Today, all pods in the Kuadrant deployment accept ingress traffic from any source in the cluster.
+Any compromised pod or misconfigured workload can reach internal gRPC, wasm-serving, metrics, and OIDC endpoints without restriction.
+There is no defense-in-depth at the network layer.
+
+This is a security gap that needs to be closed.
+The feature establishes a deny-by-default ingress posture with explicit allow rules for known traffic patterns.
+
+Several constraints shape the approach:
+
+- **OLM compatibility**: OLM bundle policies only cover the operator — operand policies must be managed at runtime regardless.
+Additionally, not all OLM versions support NetworkPolicy in bundles.
+- **Operand placement**: Operands (Authorino, Limitador) are deployed in the Kuadrant CR's namespace, which may differ from the operator namespace.
+Policies must follow.
+- **Egress complexity**: Locking down egress would require tracking Redis endpoints (Limitador), OIDC providers (Authorino), tracing endpoints, DNS resolution, and API server IPs — all dynamic and fragile.
+Ingress-only deny provides meaningful isolation without this complexity.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When the kuadrant-operator starts, it automatically creates NetworkPolicy resources that lock down ingress traffic to all Kuadrant components.
+No user action is required — network isolation is always on.
+
+From a user's perspective:
+
+- **Nothing changes operationally.** All existing traffic patterns (metrics scraping, Gateway-to-Authorino gRPC, Gateway-to-Limitador gRPC, wasm-shim loading, console plugin access) continue to work.
+The operator creates allow rules for all known traffic flows.
+- **Unauthorized traffic is blocked.** Pods outside the expected namespaces cannot reach Kuadrant's internal ports.
+A random pod in a user namespace cannot call Authorino's gRPC endpoint.
+Metrics ports are open to all to support any monitoring stack, but gRPC and wasm ports are restricted to known sources.
+- **Gateway namespace changes are handled automatically.** When a new Gateway is created in a namespace, the operator updates its NetworkPolicies to allow traffic from that namespace.
+When a Gateway is removed, the rules are pruned.
+- **Kuadrant CR namespace is respected.** If the Kuadrant CR (and therefore Authorino and Limitador) is in a different namespace from the operator, the operator creates appropriately scoped policies in both namespaces.
+Policies only target pods labeled `kuadrant.io/managed: "true"`, so other workloads in either namespace are unaffected.
+
+The only visible artifact of this feature is the set of `NetworkPolicy` resources in the operator and operand namespaces, all labeled with `kuadrant.io/managed: "true"`.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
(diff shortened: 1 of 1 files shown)


rfcs/0022-network-policy.md
```diff
@@ -0,0 +1,569 @@
...
+- Feature Name: `network-policy`
+- Status: Draft
+- Start Date: 2026-08-10
+- RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
+- Issue tracking: [CONNLINK-1205](https://issues.redhat.com/browse/CONNLINK-1205)
+
+# Summary
+[summary]: #summary
+
+The kuadrant-operator will dynamically create and manage standard Kubernetes NetworkPolicy objects (`networking.k8s.io/v1`) at runtime to provide ingress-only network isolation for all components across the operator and operand namespaces.
```
maleck13: So the kuadrant operator will create these. I assume that is because the kuadrant operator has the view of which gateways exist and what namespaces they are in? There is a little tension here in my mind as the kuadrant operator knows the gateway namespaces but really each individual operator knows what its operand should accept and on what port.

maleck13: After discussion, the kuadrant operator knows what ports to protect because the service of the operand is in the DAG

rfcs/0022-network-policy.md
```diff
@@ -0,0 +1,569 @@
...
+  # Metrics — open to all
+  - ports:
+    - protocol: TCP
+      port: 8080
+  # gRPC extensions — within operator namespace only
+  - from:
+    - podSelector: {}
+    ports:
+    - protocol: TCP
+      port: 50051
```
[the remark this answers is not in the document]
maleck13: This is for extension controllers to be able to communicate with the kuadrant operator via grpc. Limiting it to the same namespace is a reasonable thing at this point

rfcs/0022-network-policy.md
```diff
@@ -0,0 +1,569 @@
...
+  podSelector:
+    matchLabels:
+      app: limitador-operator
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 8080
+  policyTypes:
+  - Ingress
+```
```
maleck13: These 3 are the same bar the label selector any reason not to just add `app: dns-operator` for example and name the policy `kuadrant-controlplane-metrics`?

rfcs/0022-network-policy.md
```diff
@@ -0,0 +1,569 @@
...
+- **Operand namespace**: Contains Authorino and Limitador deployments.
+Determined by where the user creates the Kuadrant CR.
+
+Either namespace may contain workloads not managed by Kuadrant.
+To avoid disrupting those workloads, both the operator and operand namespaces use a label-scoped deny model targeting only pods with `kuadrant.io/managed: "true"`.
+
+| Scope | Deny Policy | Rationale |
+|---|---|---|
+| Operator namespace | `podSelector: { kuadrant.io/managed: "true" }` (label-scoped) | Other workloads may share this namespace. |
+| Operand namespace (if different) | `podSelector: { kuadrant.io/managed: "true" }` (label-scoped) | Other workloads may share this namespace. |
```
maleck13: while I see this as a nice way to target. I wonder is it strictly necessary or is there a set of labels that could be used that already exist?

rfcs/0022-network-policy.md
```diff
@@ -0,0 +1,569 @@
...
+**Dynamic gateway namespace handling:** When a new Gateway is created in a previously unseen namespace, the next reconciliation adds that namespace to the ingress `from` rules on the kuadrant-operator (port 8082), authorino (ports 50051, 5001), and limitador (port 8081) policies.
+When a Gateway is removed from a namespace (and no other Gateways remain there), those namespace entries are pruned from the ingress rules.
+
+**Operand namespace changes:** If the Kuadrant CR is deleted from one namespace and created in another, the reconciler removes the operand policies (`kuadrant-deny-managed`, `kuadrant-np-authorino`, `kuadrant-np-limitador`) from the old namespace and creates them in the new one.
+
+## Lifecycle and Cleanup
+
+NetworkPolicies use Kubernetes owner references for lifecycle management:
+
+- **Operator namespace policies**: Owned by the kuadrant-operator Deployment.
```
maleck13: poss worth looking at @mikenairn RFC that introduces a KuadrantControlPlane CR

rfcs/0022-network-policy.md
```diff
@@ -0,0 +1,569 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- **Labeling prerequisite**: All Kuadrant component pods must carry the `kuadrant.io/managed: "true"` label for the deny policy to cover them.
+Any component pod missing this label will not be selected by the deny policy and will remain open to all ingress.
+This labeling work must be completed across all sub-operators before network policies are effective.
+- **No egress restrictions** means a compromised pod can still make outbound connections to arbitrary destinations.
+This is an accepted trade-off for the significant reduction in implementation complexity.
+- **Pod selector labels** (`app: kuadrant-operator`, `app: authorino`, etc.) must match the actual labels on component deployments.
```
maleck13: If the kuadrant operator has the deployment and service in the DAG, it could ensure that the policy it is creating uses an existing label rather than relying on a new one?

maleck13 (commented):
Generally this looks good. 

Main concern was around whose responsibility it is to create the network policy. Normally it would be the job of the operator responsible for the operand. That said as the kuadrant operator has the view of the gateways and also has a view of the ports (due to the DAG) it has all the info needed and so doing this in one place rather than individually doing it in each operator makes sense especially as the deployment model supported is via the Kuadrant operator rather than individual operators. 

The other concern is around the generic label kuadrant.io/managed: true, This may require additional work, if we have a view of the podTemplate via the deployments, we could dynamically figure out what label to use based on those available favouring the managed: true label?  Not against the approach just thinking if there are existing labels we could use.

Finally for MCP Gateway, its objects are not in the DAG so we will need to create and manage those in the controller . I have linked a related issue

rfcs/0022-network-policy.md
Boomatang: This is a very much nice to have, and the more I dig into the actually implementation I am finding out that the labels are all over the place. It is looking like we need to do this dynamically.

rfcs/0022-network-policy.md
Boomatang: yes we can do something like that. It would have being not to need to.

rfcs/0022-network-policy.md
Boomatang: I had a look over https://github.com/Kuadrant/architecture/blob/main/rfcs/0019-olmv1-operator-consolidation.md and I did see anything in it with would conflict with what we are planning. Is there any sections that you are concerned over.

rfcs/0022-network-policy.md
Boomatang: I did think about this, but the ports can be configured per deployment. Also having the configurations separate allows us to make use of the ownerReference. If the deployment for the dns-operator is remove k8s also removes the network policy related to it.

